PR #65, opened 2022-12-31
better env management

### LINKED ISSUE #55: better .gitignore for .env
Follow up from PR. We're relying on a weird git corner case to protect our `.env` files

ann-kilzer:
Resolves #55 

Better env management

Testing instructions:

try editing `env.sample` and ensure git sees the changes


ermish (approved):
LGTM!

PR #67, opened 2022-12-31
Dockerize Postgres

### LINKED ISSUE #66: dockerize the database
Follow up from #27

ann-kilzer:
Resolves #66 👹 

- [x] run postgres in docker
- [x] get the migrations to run
- [x] Ensure migrations run from outside the container, e.g. `yarn migrate`

## Testing instructions

1. Check out the readme changes and follow the instructions to see if you get a container working
2. Run `yarn migrate` from your localhost (not inside docker) to ensure there is a seamless connection between your local environment and the dockerized db
3. Try `yarn prisma studio` to ensure there is a seamless connection between the explorer and the dockerized db


.env.sample
```diff
@@ -3,6 +3,7 @@ POSTGRES_PASSWORD=
 POSTGRES_HOST=127.0.0.1
 POSTGRES_DB=findadoc
 SHADOW_DB=shadow
+POSTGRES_PORT=5400
```
ann-kilzer, line 6: Need to clarify between baremetal (5432) and docker

docker/init-user-db.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+# load env variables
+source ./.env
+
+echo "😃😃😃😃😃😃😃😃😃😃😃😃😃😃😃😃😃😃😃😃😃😃😃😃😃"
```
ann-kilzer, line 6: I don't think this file is actually running

docker/db-init.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+echo "😃😃😃😃😃😃DB SETUP TIME😃😃😃😃😃"
+
+echo $POSTGRES_DB
+
+psql -U $POSTGRES_USER -p $POSTGRES_PORT -W $POSTGRES_PASSWORD << EOF
```
ann-kilzer, line 7: This is failing

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+    networks:
+      new:
+        aliases:
+          - database
+    env_file:
+      - .env
+    entrypoint: /docker/db-init.sh
+    healthcheck:
+      test:
+        - postgres psql -c "SELECT datname FROM pg_catalog.pg_database WHERE datname='${SHADOW_DB}'"
```
ann-kilzer, line 34: Unsure if this healthcheck is working

RageZBla:
@ann-kilzer few things:

- `docker-compose` super unintuitive behavior `depends_on` would not do anything but start the container in whatever order, it does not wait for the service to be healthy. [reference](https://docs.docker.com/compose/startup-order/). Because of this behavior, you are forced to build some safe check to wait for things to be available
-  The network layers of docker allow exposing things on the local machine on different ports, which means that from inside the container `database:5432`(default Postgres port) would be available. On the other hand, it would be mapped to port 5400 on your host OS. 
- more of a question: do you want to use docker for everything or just the database part? I was not sure.

docker/db-init.sh
```diff
@@ -1,12 +1,26 @@
...
+    PGPASSWORD=$POSTGRES_PASSWORD psql -U $POSTGRES_USER -p $POSTGRES_PORT -h database -tc "SELECT 1 FROM pg_database WHERE datname = '$db'" | grep -q 1 || \
+    PGPASSWORD=$POSTGRES_PASSWORD psql -U $POSTGRES_USER -p $POSTGRES_PORT -h database << EOF
+       CREATE DATABASE $db;
+       GRANT ALL PRIVILEGES ON DATABASE $db TO $POSTGRES_USER;
+EOF
+}
+
+create_database_if_not_exists $POSTGRES_DB
+create_database_if_not_exists $SHADOW_DB
+touch out/db-init-done
```
ann-kilzer, line 26: nice

docker-compose.yml
```diff
@@ -11,14 +11,13 @@ services:
...
-      test:
-        - pg_isready -U ${POSTGRES_USER} -d ${POSTGRES_DB}
+      test: ["CMD", "pg_isready -U ${POSTGRES_USER} -d ${POSTGRES_DB}"]
       interval: 10s
       timeout: 3s
       retries: 5
     volumes:
       - pgdata:/var/lib/postgresql/findadoc-data # persist data even if container shuts down
-  db-cofigure:
+  db-configure:
```
ann-kilzer, line 20: 😆

docker-compose.yml
```diff
@@ -0,0 +1,76 @@
...
+    volumes:
+      - .:/target
+    command: "yarn install"
+  migrate:
+    depends_on:
+      - db-configure
+      - install-deps
+    env_file:
+      - .env
+    environment:
```
ann-kilzer, line 56: I'm overriding the URLs for prisma's connection inside docker vs. outside

.env.sample
```diff
@@ -1,8 +1,9 @@
 POSTGRES_USER=postgres
 POSTGRES_PASSWORD=
-POSTGRES_HOST=127.0.0.1
 POSTGRES_DB=findadoc
 SHADOW_DB=shadow
+POSTGRES_HOST=findadoc-server-db-1
```
ann-kilzer, line 5: 🤔 this is not the right path

RageZBla: @ann-kilzer nope, from inside thé containers it would be `db` or `database`

ann-kilzer: Here I am configuring it for outside the containers. I override the value inside the docker compose file

ann-kilzer: E.g. I need this value to be the outside-the-container path

docker/db-init.sh
```diff
@@ -0,0 +1,26 @@
...
+until PGPASSWORD=$POSTGRES_PASSWORD psql -h ${DOCKER_POSTGRES_HOST} -U $POSTGRES_USER -p $DOCKER_INTERNAL_PORT -c '\q'; do
+  >&2 echo "Postgres is unavailable - sleeping 😴"
+  sleep 1
+done
+  
+>&2 echo "Postgres is up 🙌 - executing command"
+
+function create_database_if_not_exists() {
+    local db="$1"
+    PGPASSWORD=$POSTGRES_PASSWORD psql -U $POSTGRES_USER -h database -tc "SELECT 1 FROM pg_database WHERE datname = '$db'" | grep -q 1 || \
```
RageZBla, line 17: @ann-kilzer nit: maybe worth to use the environment variable for the hostname.

RageZBla (approved):

README.md
```diff
@@ -28,32 +28,24 @@ yarn prepare
...
-createuser -s postgres
-brew services restart postgresql
-```
 
-1. Init the database
-```sh
-./init-db.sh
-```
+For simplicity we provide a dockerized postgres. As a prerequisite, you will need to install 
+[Docker Desktop](https://www.docker.com/products/docker-desktop/).
```
ermish, line 33: future ticket: thinking about a new contributor, perhaps we should make a prerequisites section that highlights different tools the  dev needs to install to contribute. (I think it's yarn 2, node 18, and docker desktop at this point 🤔 )

docker-compose.yml
```diff
@@ -0,0 +1,79 @@
+version: "3.8"
+services:
+  db:
+    image: postgres:14-alpine
+    networks:
+      new:
+        aliases:
+          - database
+    env_file:
```
ermish, line 9: is there a postgres option to have all persisted data encrypted?

docker-compose.yml
```diff
@@ -0,0 +1,79 @@
...
+    links:
+      - db
+    networks:
+      - new
+    image: node:18.12.1-alpine
+    working_dir: /target
+    volumes:
+      - .:/target
+      - status:/in
+    command: "sh -c 'until [[ -f /in/db-init-done && -d /target/node_modules ]]; do sleep 1; done; yarn migrate; yarn prisma db seed'"
```
ermish, line 71: idea: it seems like we're doing a fair amount of setup in the compose script. It might be a little cleaner to build a wrapper docker image around the `postgres:14-alpine` image and move things like this command inside the container. 
This would lessen our dependency on docker compose and make the move easier to something like AWS container services easier in the future.

ermish: non-blocking question: ^ after I was looking at the different services in the compose file, should `db-configure`, `install-deps`, and `migrate` all be tasks under a singular `db` docker file? 

I was imagining this compose file is going to managing our node service, db, etc and everything for the findadoc backend. Is this what you're thinking or rather a separate compose file for the db and node services?

ermish, line 71: potentially blocking question or new ticket: is this only meant for a creating local database and idempotent? or are the migrations and seeding also going to run in prod?

I'm guessing the seeding is only meant to be conditionally ran for a new db and not prod/staging envs. We might want this to be a `.env` variable passed in to run seed data.

docker/db-init.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+set -euo pipefail
+echo "😃😃😃😃😃😃 DB SETUP TIME 😃😃😃😃😃"
+
+echo $POSTGRES_DB
+
+
+until PGPASSWORD=$POSTGRES_PASSWORD psql -h ${DOCKER_POSTGRES_HOST} -U $POSTGRES_USER -p $DOCKER_INTERNAL_PORT -c '\q'; do
```
ermish, line 8: new issue idea: Could this all be moved into the `db` docker container? 
I usually see these kind of scripts within the container

docker-compose.yml
ann-kilzer: Hmm definitely something to look into in the future but I don't know

docker-compose.yml
ann-kilzer: Honestly this took several weekends to get working and I don't want to waste energy optimizing it when it works. Fine for folks to tune up later

docker/db-init.sh
ann-kilzer: See my other comment. Hypothetically it could, but I don't really see it bringing a lot of value vs. getting mvp working

docker-compose.yml
ann-kilzer: Seeding will be done in production. Im making a special dev seeding step on a branch today

ermish (approved):
LGTM

PR #69, opened 2022-12-31
fix codegen to add missing types

theyokohamalife:
There was a plugin missing in the `typesgeneratorconfig.ts` that was preventing all of the necessary types from being generated. This fix will allow us to change all of the types that are set to `any` to their appropriate types.

## How to test

1. Delete the `src/typeDefs/gqlTypes.ts` file and run the following:

```
yarn dev
yarn generate
```

2. Check that the generated file includes the `export type` and `export enum` that are missing from `main`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "homepage": "https://github.com/ourjapanlife/findadoc-server#readme",
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


ermish (approved):
Looks good! Wow was that all that was missing? nice find :shipit:

PR #72, opened 2023-01-01
Simplify migrations

### LINKED ISSUE #71: Simplify migrations
Since we're not deployed to prod / staging yet, let's neaten up the DB migrations

ann-kilzer:
Resolves #71 

We won't do this often but it's safe because we don't have a production or staging DB stood up. Neaten up the initial DB migration for clarity.

## Testing: 

Run `yarn prisma migrate` locally (assumes DB setup from the readme is complete)


prisma/migrations/20230101042848_init/migration.sql
```diff
@@ -1,6 +1,9 @@
 -- CreateTable
```
ann-kilzer, line 1: Reminder that this file is generated from `prisma.schema`

ermish (approved):
LGTM

PR #73, opened 2023-01-01
Seed data for development environment

### LINKED ISSUE #52: Database Seed data
follow-up from #27 #50 

Add Seed data to the database for known entities like SpokenLanguage, Degree, Specialty

ann-kilzer:
Resolves #52 

Add some seed data for dev only

## Testing

1. `yarn prisma migrate reset` should reset and reseed the database
2. Examine the schema with `yarn prisma studio`


prisma/devSetup.ts
```diff
@@ -0,0 +1,193 @@
+/* eslint-disable func-names */
+/* eslint-disable no-console */
+import { PrismaClient } from '@prisma/client';
+import loadCSVFromFile from './loadCsv';
+
+const seedHealthcareProfessionals = async function (prisma: PrismaClient, verbose = false) {
```
ann-kilzer, line 6: Data like professionals, specialties, and degrees should be seeded into all environments.

prisma/schema.prisma
```diff
@@ -85,9 +85,9 @@ model HealthcareProfessional {
   HealthcareProfessionalDegree    HealthcareProfessionalDegree[]
   spokenLanguages                 HealthcareProfessionalSpokenLanguage[]
 
-  contact Contact @relation(fields: [contactId], references: [id])
+  contact Contact? @relation(fields: [contactId], references: [id])
```
ann-kilzer, line 88: As I've worked with the data I am lightening up some of the requirements to make it easier to use

prisma/seed.ts
```diff
@@ -3,27 +3,11 @@
...
 // https://www.prisma.io/docs/guides/database/seed-database
 import { PrismaClient } from '@prisma/client';
-import { parse } from 'csv-parse/sync';
-import * as fs from 'fs';
+import loadCSVFromFile from './loadCSV';
+import devSetup from './devSetup';
 
 const prisma = new PrismaClient();
 
-// loads a CSV file from the filesystem, ready for parsing
```
ann-kilzer, line 11: Moved to loadCsv.ts

prisma/devSetup.ts
```diff
@@ -0,0 +1,211 @@
...
+    spokenLangList.forEach(async (lang) => {
+      const dbSpokenLang = await prisma.spokenLanguage.findFirst(
+        {
+          where: {
+            iso639_3: lang,
+          },
+        },
+      );
+      if (dbSpokenLang) {
+        // TODO change to upsert
```
ann-kilzer, line 78: It would be better if I could make this an upsert so that running the seed file multiple times does not result in duplicate data

prisma/devSetup.ts
```diff
@@ -0,0 +1,211 @@
...
+  const lastJa = 5;
+  const degreeCol = 6;
+  const specialtyCol = 7;
+  const spokenLangCol = 8;
+
+  const devData:string[][] = loadCSVFromFile('./prisma/seedData/devHealthcareProfessionals.csv');
+
+  devData.forEach(async (row: string[], index: number) => {
+    // generate some unique ids using even/odd strategy
+    const enID = 2 * index;
```
ermish, line 21: future issue: before we go to production, it would be a good idea to use non-predictable ids for security purposes.

prisma/devSetup.ts
```diff
@@ -0,0 +1,211 @@
...
+              healthcareProfessionalId: newHealthPro.id,
+            },
+          },
+        );
+      }
+    });
+
+    // Link Degrees
+    const degreeList = row[degreeCol].split(',');
+    degreeList.forEach(async (degree) => {
```
ermish, line 92: bug: in node, `forEach(async` doesn't work (unless they recently fixed it.) it doesn't guarantee that every iteration is completed before the next. It launches them all essentially in parallel. 

Personally, I've always seen helper functions for each company or I've created one myself. Instead of `foreach(async` you can use `await Promise.all(degreeList.map(async` or a helper function if you want to make one

prisma/loadCsv.ts
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable no-console */
+import { parse } from 'csv-parse/sync';
+import * as fs from 'fs';
+
+// loads a CSV file from the filesystem, ready for parsing
+export default function loadCSVFromFile(filename: string, discardHeader = true, delimiter = ',') {
```
ermish, line 6: nit: could make this an async function. not a big deal since it's just for our seed data.

prisma/migrations/20230101052800_physical_address/migration.sql
```diff
@@ -0,0 +1,30 @@
+/*
+  Warnings:
+
+  - Added the required column `physicalAddressId` to the `Contact` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "Contact" ADD COLUMN     "physicalAddressId" INTEGER NOT NULL;
```
ermish, line 8: nit: spacing.

prisma/migrations/20230101095749_simplify_provider/migration.sql
```diff
@@ -0,0 +1,8 @@
+-- DropForeignKey
+ALTER TABLE "HealthcareProfessional" DROP CONSTRAINT "HealthcareProfessional_contactId_fkey";
+
+-- AlterTable
+ALTER TABLE "HealthcareProfessional" ALTER COLUMN "contactId" DROP NOT NULL;
```
ermish, line 5: question: is this just for updating our current data or is this meant to be run every time?

prisma/migrations/20230101120731_name_change/migration.sql
```diff
@@ -0,0 +1,16 @@
...
+
+  - You are about to drop the column `physicalAddressId` on the `Contact` table. All the data in the column will be lost.
+  - Added the required column `addressId` to the `Contact` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- DropForeignKey
+ALTER TABLE "Contact" DROP CONSTRAINT "Contact_physicalAddressId_fkey";
+
+-- AlterTable
+ALTER TABLE "Contact" DROP COLUMN "physicalAddressId",
```
ermish, line 12: question: is this just for updating our current data or is this meant to be run every time?

prisma/migrations/20230101095749_simplify_provider/migration.sql
ermish: should this be `DROP COLUMN` instead of alter?

prisma/devSetup.ts
ermish: ^ I'll approve so we can keep going and if you want , we can fix this in an iterative PR

ermish (approved):
LGTM. added one bug that might be worth updating now

prisma/devSetup.ts
ermish: woah! just learned that ES2018 there's now a `for await` simplified option! I'm excited to try this 🤩 ✨ 
https://stackoverflow.com/a/50874507/3417143

prisma/devSetup.ts
ann-kilzer: Ahh, good thought. Do you think this matters for non-PII data like degrees, specialities, etc?

prisma/devSetup.ts
ann-kilzer: I actually learned this while running this, with the counter defined outside the loop and saw there was a race condition, so I should update these to use the `index` inside the loop...

prisma/migrations/20230101052800_physical_address/migration.sql
ann-kilzer: This is autogenerated ... not sure why it's doing that but  ¯\_(ツ)_/¯

prisma/migrations/20230101095749_simplify_provider/migration.sql
ann-kilzer: This is generated by Prisma. I'll probably squash these migrations together so take a look at the next PR to see if you're happy with it.

prisma/migrations/20230101120731_name_change/migration.sql
ann-kilzer: Migrations are only run once on the database

prisma/loadCsv.ts
ann-kilzer: 🤔 I took a look, but which contained methods are async? `readFileSync` gives me a "this has no effect" And the parse functions are all synchronous

prisma/devSetup.ts
ermish: for privacy, no.....we should be good. 👍  Foreign keys to that data will be with the PII data that should be encrypted like healthcare professionals, facilities, etc..  There's no risk/merit of black hat usage.

prisma/loadCsv.ts
ermish: I would switch `readFileSync` to `fs.promises.readFile`. 
Generally, my recommendation is to `await`/ promisify any external dependency or slow, complex function 🔢

prisma/migrations/20230101052800_physical_address/migration.sql
ermish: ha! that's interesting XD I bet it's a prism dev thing

prisma/migrations/20230101120731_name_change/migration.sql
ermish: ohhhhh ok....that makes sense. I was thinking these were just interim migrations and then long-term migrations were idempotent. Ryokai 👍

PR #75, opened 2023-01-02
Docker bugfix

ann-kilzer:
I found a small bug after merging seeding + docker due to OS X ignoring casing 😳 

## Testing

`docker compose up` should not have errors


ermish (approved):
LGTM :shipit: 🚀

PR #92, opened 2023-01-12
fix: set platform version to make prisma compatible with M1

### LINKED ISSUE #90: Prisma ARM error
We noticed an error when running `docker compose up` on an ARM / M1 mac.

## Error Message

```
 Environment variables loaded from .env
findadoc-server-migrate-1       | Error: Unable to establish a connection to query-engine-node-api library. It seems that the current architecture arm64 is not supported, or that libc is missing from the system.
findadoc-server-migrate-1       | Details: Unable to require(`/target/node_modules/@prisma/engines/libquery_engine-linux-arm64-openssl-3.0.x.so.node`)
findadoc-server-migrate-1       |  Error loading shared library ld-linux-aarch64.so.1: No such file or directory (needed by /target/node_modules/@prisma/engines/libquery_engine-linux-arm64-openssl-3.0.x.so.node)
findadoc-server-migrate-1       | [Context: getDmmf]
```

theyokohamalife:
Resolves #90 

# What changed
1. Adds `platform: linux/amd64` to the Docker build.
2. Updates the README steps for getting the database running

# Testing instructions
1. Follow the updated steps in the README to get the database and docker running.
2. Do the following query in GraphQL studio:

Operation:
```
query Query($iso6393: String!) {
  spokenLanguage(iso639_3: $iso6393) {
    nameNative
    nameJa
    nameEn
    iso639_3
  }
}
```

Variables
```
{
  "iso6393": "eng"
}
```

### Screenshot
Screenshot 2023-01-13 at 9 06 41 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,15 +53,14 @@ Edit the .env file and set the database password
 docker compose up
 ```
 
-3. Run the migrations
-```sh
-yarn prisma migrate dev
-```
+### Database and ORM commands
 
-4. Run the seeds
-```sh
-yarn prisma db seed
-```
+Re-run migrations with `yarn prisma migrate dev`
+
+Create a new migration by:
+1. Editing schema.prisma and save
+2. Run `yarn prisma migrate  dev --name snake_case_title`
+3. Before running the backend, regenerate the Prisma Client with `yarn prisma generate`
 
 ### Note about migration files
 
--- a/README.md
+++ b/README.md
@@ -53,6 +53,16 @@ Edit the .env file and set the database password
 docker compose up
 ```
 
+3. Run the migrations
+```sh
+yarn prisma migrate dev
+```
+
+4. Run the seeds
+```sh
+yarn prisma db seed
+```
+
 ### Note about migration files
 
 Migration files should never edited after merging to `main` in order to keep an accurate database history. If you want
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -49,7 +49,7 @@ services:
     working_dir: /target
     volumes:
       - .:/target
-    command: "yarn install"
+    command: "yarn install; yarn prisma generate"
   migrate:
     depends_on:
       - db-configure
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: "3.8"
+version: '3.8'
 services:
   db:
     image: postgres:14-alpine
@@ -11,11 +11,7 @@ services:
     ports:
       - ${POSTGRES_PORT}:${DOCKER_INTERNAL_PORT}
     healthcheck:
-      test:
-        [
-          "CMD",
-          "pg_isready -U ${POSTGRES_USER} -d ${POSTGRES_DB}"
-        ]
+      test: ['CMD', 'pg_isready -U ${POSTGRES_USER} -d ${POSTGRES_DB}']
       interval: 10s
       timeout: 3s
       retries: 5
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -53,6 +53,16 @@ Edit the .env file and set the database password
...
+3. Run the migrations
+```sh
+yarn prisma migrate dev
+```
+
+4. Run the seeds
+```sh
+yarn prisma db seed
+```
+
```
ann-kilzer, line 65: Docker compose does migrations and seeding this internally so you don't need to run it again as part of the Database setup.

Can we instead make a section like so
```suggestion
### Database and ORM commands

Re-run migrations with `yarn prisma migrate dev`

Create a new migration by:
1. Editing schema.prisma and save
2. Run `yarn prisma migrate  dev --name snake_case_title`
3. Before running the backend, regenerate the Prisma Client with `yarn prisma generate`

```

Sadly I can't do triple-backticks in the suggestion section or Markdown has a parsing fit, but feel free to adjust further if you like.

Lastly, seeding shouldn't be something we regularly need to run so probably fine to skip here as it's dockerized.

ann-kilzer (commented):
Looking nice, would like the README adjusted for clarity

theyokohamalife:
@ann-kilzer  I found this info from the Prima issues regarding M1 macs: https://github.com/prisma/prisma/issues/8478#issuecomment-1355209706

ann-kilzer:
https://github.com/ourjapanlife/findadoc-server/issues/90#issuecomment-1382041787

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: "3.8"
+version: '3.8'
```
ann-kilzer, line 1: nit: are we accidentally running a formatter on this? 

Either quote style is fine, but it's better to not mix in arbitrary style changes.

docker/db-init.sh
```diff
@@ -3,6 +3,7 @@ set -euo pipefail
 echo "😃😃😃😃😃😃 DB SETUP TIME 😃😃😃😃😃"
 
 echo $POSTGRES_DB
+uname -a
```
ann-kilzer, line 6: Oh this was just a test 😆 fine to remove
```suggestion
```

ann-kilzer (commented):
The changes here don't appear to have anything to do with node. Can you please update the ticket description and title accordingly?

It looks like the platform addition may be making things work

docker-compose.yml
ann-kilzer: (Since we're not using prettier in this repo, it may be a good idea to _turn off_ Prettier in VS Code for this workspace)

ann-kilzer:
I confirmed that docker runs the containers on an Intel mac using this code. Instead of firing up the backend, I tested with `yarn prisma studio` to quickly inspect the database.

Let me know when you have a clearer description of the changes and I can approve it.

docker-compose.yml
```diff
@@ -43,13 +39,14 @@ services:
       - db-configure
     env_file:
       - .env
+    platform: linux/amd64
     image: node:18.12.1-alpine
```
ann-kilzer, line 43: @RageZBla I understand @theyokohamalife was able to get this working locally by running `docker rmi node:18.12.1-alpine` on her machine, but I still see the node alpine image listed here. Can you please check this so we have a repeatable docker build?

docker-compose.yml
theyokohamalife: I have Prettier turned off and ran `yarn lint --fix`. It still uses single quotes. 🤷🏾‍♀️

ann-kilzer: Eslint does not run on .yml files, I am guessing an editor save did it.

ann-kilzer (approved):
Great! 🏎️

Issue #87, opened 2023-01-08
Connect SQL to GraphQL


ann-kilzer:
# Requirements

- [x] Connect API layer to DB layer using prisma ORM
- [x] use the seeded DB data instead of other static data

# Dependencies

- None (Add any issue dependencies here)

PR #97, opened 2023-01-14
Connect DB and API

ann-kilzer:
Part of #87 

So there are some big changes here, let me know if you want to chat to square things up, especially regarding the schema changes.

# What changed
- [x] Connecting the SQL layer to the GraphQL
- [x] Remove mockData
- [x] Adjust some of the GraphQL models to match the DB data (I'll make comments inline so we can discuss)
- [x] Adjust Specialty DB to match GraphQL
- 🏉  Link healthcare professional to facility => another PR
- [x] healthcare professional (simplified version)

# Testing instructions

Sample queries for Apollo explorer are provided:

## Specialties

```graphql
query Specialties {
  specialties {
    id
    names {
      locale
      name
    }
  }
}
```

## Specialty by Id
```graphql
query Specialties($specialtyId: ID!) {
  specialty(id: $specialtyId) {
    id
    names {
      locale
      name
    }
  }
}

{
  "specialtyId": 1
}
```

## Facilities (missing HealthcareProfessional)

```graphql
query Facilities{
  facilities {
    contact {
      email
      mapsLink
      phone
      website
    }
    id
    nameEn
    nameJa
  }
}
```

## Facility by ID (missing HealthcareProfessional)

```graphql
query Facilities($facilityId: ID!){
  facility(id: $facilityId) {
    contact {
      email
      mapsLink
      phone
      website
    }
    id
    nameEn
    nameJa
  }

{
  "facilityId": 1
}
}
```

## HealthcareProfessional (simplified)

```graphql
query HealthcarePro{
  healthcareProfessionals {
    id
    names {
      firstName
      lastName
      locale
      middleName
    }
  }
}
```

## HealthcareProfessional by ID (simplified)

```graphql
($healthcareProfessionalId: ID!){
  healthcareProfessional(id: $healthcareProfessionalId) {
    names {
      firstName
      lastName
      locale
      middleName
    }
    id
  }
}

{
  "healthcareProfessionalId": 1
}
```


src/typeDefs/schema.graphql
```diff
@@ -7,16 +7,12 @@ type Contact {
 
 type Facility {
   id: ID!
-  names: [FacilityName]!
```
ann-kilzer, line 10: I simplified the facility names on the DB end and am adjusting the GraphQL to match.

My reasoning: Most medical websites in Japan have both a Japanese and an English or Romaji name. I haven't seen clinics using, say, Vietnamese names, etc., and I thought an overly complex model would be harder for MVP. Might be my own bias, and checking with Russ 🤔

src/typeDefs/schema.graphql
```diff
@@ -49,22 +45,14 @@ type PersonName {
...
 
 input SpecialtyInput {
-  names: [SpecialtyNameInput]
+  nameEn: String
+  nameJa: String
 }
 
 type Specialty {
   id: ID!
-  names: [SpecialtyName]
```
ann-kilzer, line 57: I should change this back... 🤔 we'll definitely want Specialties to have many translations 👀

prisma/migrations/20230114114951_specialties/migration.sql
```diff
@@ -0,0 +1,23 @@
+/*
```
ann-kilzer, line 1: I'm updating the database here to better match the GraphQL

prisma/schema.prisma
```diff
@@ -12,18 +12,7 @@ datasource db {
   shadowDatabaseUrl = env("SHADOW_DB_URL")
 }
 
-// Represents the name of a human individual in both Japanese and Romanji characters.
```
ann-kilzer, line 15: My original data model for names had an unnecessary level of indirection, and I simplified it to work better with the GraphQL

prisma/schema.prisma
```diff
@@ -59,13 +47,23 @@ model SpokenLanguage {
 
 // Represents the area a professional works in, such as Gynecology, gastroenterology, oncology, physical therapy, etc.
 model Specialty {
-  id     Int    @id @default(autoincrement())
-  nameJa String
-  nameEn String
+  id    Int             @id @default(autoincrement())
```
ann-kilzer, line 50: Model expanded to match the GraphQL

src/services/facilityService.ts
```diff
@@ -1,13 +1,48 @@
...
+    contact: PrismaContact;
+})
+
+function convertPrismaToGraphQLFacility(input: FacilityAndRelations) {
+    return {
+        id: String(input.id),
+        nameJa: input.nameJa,
+        nameEn: input.nameEn,
+        contact: input.contact,
+        healthcareProfessionals: []
```
ann-kilzer, line 16: hydrating these objects is a bit complex. I'd love to work more on healthcareProfessionals in a following PR

prisma/migrations/20230114114951_specialties/migration.sql
ann-kilzer, line 1: Reminder that migrations are generated files and will not be hand-edited. Please see `schema.prisma` for a clearer view of the changes

prisma/migrations/20230114134913_simplify_name/migration.sql
```diff
@@ -0,0 +1,40 @@
+/*
```
ann-kilzer, line 1: Reminder that migrations are generated files and will not be hand-edited. Please see `schema.prisma` for a clearer view of the changes

src/typeDefs/schema.graphql
```diff
@@ -27,21 +23,21 @@ input HealthcareProfessionalInput {
...
   id: ID!
-  names: [PersonName]!
+  names: [LocaleName]!
   degrees: [Degree]!
   spokenLanguages: [SpokenLanguage]!
   specialties: [Specialty]!
   acceptedInsuranceOptions: [Insurance]!
 }
 
-input PersonNameInput {
```
ann-kilzer, line 37: Keeping in sync with the DB naming

prisma/schema.prisma
ermish: 💯  this makes more sense to me

prisma/schema.prisma
```diff
@@ -59,13 +47,23 @@ model SpokenLanguage {
...
+  names SpecialtyName[]
 
   HealthcareProfessionalSpecialty HealthcareProfessionalSpecialty[]
 }
 
+model SpecialtyName {
```
ermish, line 56: can we use `LocaleName` instead of SpecialtyName? Or does it need distinct fields that are different?

prisma/seed.ts
```diff
@@ -45,13 +45,17 @@ async function seedSpecialties(verbose = false) {
             where: { id: index },
             update: {},
             create: {
-                nameEn: specialty[enCol],
-                nameJa: specialty[jaCol]
+                names: {
```
ermish, line 48: 💖

src/index.ts
```diff
@@ -11,7 +11,6 @@ const server = new ApolloServer({
 const port = 3001;
 
 server.listen({ port }).then(({ url }) => {
+    // eslint-disable-next-line no-console
     console.log(`🚀  Server ready at ${url}`);
 });
-
-// add some lint
```
ermish, line 17: I can't comment on the mockData file, but we should keep this. The main use case is for running automation tests.

src/services/facilityService.ts
```diff
@@ -1,13 +1,48 @@
...
+    } as Facility;
+}
 
 // TODO: add a validation step for incoming parameters
-export const getFacilityById = (id: string) => {
-    const matchingFacility = facilities.find(
-        (location: Facility) => location.id === id
+export const getFacilityById = async (id: string) => {
+    const found = await prisma.facility.findUnique(
```
ermish, line 22: nit: could we call it something like `matchingFacility` or `facility`?

src/services/facilityService.ts
```diff
@@ -1,13 +1,48 @@
...
+        {where: {
+            id: parseInt(id)
+        },
+        include: {
+            contact: true
+        }}
     );
 
-    return matchingFacility;
+    return found ? convertPrismaToGraphQLFacility(found) : null;
```
ermish, line 31: nit optimization: you could move the null check into the function so you don't have to keep doing the null check before calling

src/services/facilityService.ts
```diff
@@ -1,13 +1,48 @@
...
 
-export const getFacilities = () => facilities;
+export const getFacilities = async () => {
+    const facilities = await prisma.facility.findMany(
+        {include: {
+            contact: true
+        }}
+    );
+
+    const ret = Array<Facility>();
```
ermish, line 41: nit: could we use more descriptive names? I couldn't figure out what `ret` was till I saw it was being returned. 

```suggestion
    const gqlFormattedFacilities = Array<Facility>();
```

src/services/facilityService.ts
```diff
@@ -1,13 +1,48 @@
...
+
+    for (let i = 0; i < facilities.length; i++) {
```
ermish, line 43: personal preference nit: could use `facilities.foreach()` instead of `for()`

src/services/healthcareProfessionalService.ts
```diff
@@ -1,13 +1,53 @@
...
+import { HealthcareProfessional, LocaleName } from '../typeDefs/gqlTypes';
+import { HealthcareProfessional as PrismaHealthcareProfessional,
+    LocaleName as PrismaLocaleName } from '@prisma/client';
+
+import prisma from '../db/client';
+
+type HealthcareProfessionalAndRelations = (PrismaHealthcareProfessional & 
+    { names: PrismaLocaleName[]})
+
+function convertPrismaToGqlHealthcareProfessional(input: HealthcareProfessionalAndRelations) {
```
ermish, line 10: idea for improving readability: not sure if we need this level of abstraction yet, but typically I put mappings into a separate folder to keep the services a bit cleaner.

src/services/healthcareProfessionalService.ts
```diff
@@ -1,13 +1,53 @@
...
+        id: String(input.id),
+        names: Array<LocaleName>(),
+        specialties: [],
+        spokenLanguages: [],
+        acceptedInsuranceOptions: [],
+        degrees: []
+    } as HealthcareProfessional;
+
+    for (let i = 0; i < input.names.length; i++) {
+        ret.names.push(input.names[i] as LocaleName);
```
ermish, line 22:
```suggestion
        ret.names?.push(input.names[i] as LocaleName);
```

src/services/healthcareProfessionalService.ts
```diff
@@ -1,13 +1,53 @@
...
+export const getHealthcareProfessionals = async () => {
+    const healthPros = await prisma.healthcareProfessional.findMany({
+        include: {
+            names: true
+        }
+    });
+
+    const ret = Array<HealthcareProfessional>();
+
+    for (let i = 0; i < healthPros.length; i++) {
```
ermish, line 49: actually, this is a perfect use case for `.map()`

```suggestion
    const gqlFormmatedEntities = healthPros.map(convertPrismaToGqlHealthcareProfessional)
```

src/services/specialtyService.ts
```diff
@@ -1,13 +1,50 @@
...
+export const getSpecialtyById = async (id: string) => {
+    const found = await prisma.specialty.findUnique({ where: {
+        id: parseInt(id)
+    },
+    include: {
+        names: true
+    }});
 
-    return matchingSpecialty;
+    if (found) {
```
ermish, line 33: readability recommendation: I recommend using the "[early return pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8)"

ermish (commented):
Overall, looks awesome! It's coming together 🚀 
I saw a few changes that might be worth taking a look at before it's merged, but no major concerns

src/services/facilityService.ts
ann-kilzer: Yeah this is just very `golang` style 😎 I can adjust

src/services/facilityService.ts
ann-kilzer: I thought you asked us to be careful with foreach 🤔 

Is there a good rule of thumb for when to use each? I know foreach isn't always run in order

src/services/healthcareProfessionalService.ts
ann-kilzer: It's a good idea, can we save for a future refactor?

src/services/facilityService.ts
ann-kilzer: sure

src/services/facilityService.ts
ann-kilzer: that's a good idea and now I'm wondering why i didn't think of it 🦞

src/services/facilityService.ts
ann-kilzer: I'm going to try to find a happy medium between the golang `r` and `ret` and the Java `thisIsATemporaryVariableCounter` 

how about `gqlFacilites`

src/services/healthcareProfessionalService.ts
ann-kilzer: Not sure if I follow, and some of the refactor suggestions are conflicting. I am gonna push some of the changes and let you comment again.

src/services/specialtyService.ts
ann-kilzer: I think that's a great pattern when there are nested ifs -- to break the logic out of nests.

But in this case, seems a bit "six of one, half-dozen of the other" 🤔

ann-kilzer: Anyway it's about to disappear as part of another refactor

prisma/schema.prisma
ermish: oh the middle name and last is a good point. I'm sold 💰

src/index.ts
ermish: yep! ha, I think our first step is to start having some nice integration tests that actually use it!

src/services/facilityService.ts
ermish: 💯 sounds great to me!

src/services/facilityService.ts
ermish: sure! 

### use foreach when:
1) you want to do an operation on each record
2) order doesn't matter

### don't use foreach when:
1) you are changing the shape of the data or changing the order or size. You can use map, reduce, etc. 
2) the operation is async
3) you have 10,000+ records. It does have a minor performance hit on huge data sets

src/services/healthcareProfessionalService.ts
ermish: yup! whenever you feel like it's starting to make the files long and hard to follow is probably the right time 😸

src/services/healthcareProfessionalService.ts
ermish: ok! I'll add an example to the new changes

src/services/specialtyService.ts
```diff
@@ -1,13 +1,50 @@
...
 
-export const getSpecialties = () => specialties;
+export const getSpecialties = async () => {
+    const dbSpecialties = await prisma.specialty.findMany({ include: {
+        names: true
+    }});
+
+    const gqlSpecialties = Array<Specialty>();
+
+    dbSpecialties.forEach(specialty => {
```
ermish, line 44: Here's the example for how you could use `reduce()` here instead. Originally, you could use map to just change each record, but since it's conditionally adding, reduce makes more sense

```suggestion
    const gqlSpecialties  = dbSpecialties.reduce((convertedSpecialties, specialty) => {
        const gqlSpecialty = convertPrismaToGraphQLSpecialty(specialty);
         if (gqlSpecialty) { convertedSpecialties.push(gqlSpecialty); }         
         
         return convertedSpecialties;
}, []);
```

ermish: ^ this is pretty much how most people do array conversions, filtering, etc nowadays. It's effectively the same functionality though, so this is totally up to you which way you prefer

ermish (approved):
LGTM! nothing blocking. take a look at the reduce example and see if it's something you'd want to update here

src/services/specialtyService.ts
ann-kilzer: I think it's cool and I'd like to study it more, but would feel better when I have better test coverage... right now nothing 😭 

How about I bookmark and return when there are unit or integration tests?

PR #98, opened 2023-01-20
add editor config and adjust indentation to 4

### LINKED ISSUE #58: add editorconfig
- Add editorconfig from web project

ann-kilzer:
Resolves #58 

# What changed
- Add editor config 
- Update indentation style to be 4 as much as possible
(we can adjust the graphql and yaml files later but I am worried about merge conflicts right now

# Testing instructions
EditorConfig works automatically with many tools: https://editorconfig.org/#pre-installed
If you have VS Code, you will need to install a plugin: https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

Now when you edit files, you should automatically get the right indentation when you press tab, have the correct line endings `\n` (unix) not `CRLF` (windows). It can help us easily work with folks on different platforms. 

Thanks to @ermish for suggesting it

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,32 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+
+# Matches multiple files with brace expansion notation
+[*.{js,ts,json}]
+charset = utf-8
+
+[*.graphql]
+charset = utf-8
+indent_size = 2
+
+[*.yml]
+charset = utf-8
+indent_size = 2
+
+[*.prisma]
+charset = utf-8
+
+[*.sh]
+charset = utf-8
+
+[*.csv]
+charset = utf-8
+insert_final_newline = false
--- a/package.json
+++ b/package.json
@@ -1,64 +1,64 @@
 {
-  "name": "findadoc-server",
-  "version": "1.0.0",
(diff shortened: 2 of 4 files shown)


.editorconfig
```diff
@@ -0,0 +1,32 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+end_of_line = lf
```
ann-kilzer, line 7: 🤔 As a separate PR, it might be good to configure git to not convert `\n` to `\n\r`on windows via a `.gitattributes` file.

I have noticed that this can help Windows and Mac users develop more efficiently together, and avoid mysterious errors when ultimately running the code in a linux container. @ermish, since you use Windows, what do you think?

ermish: I like the idea, but I'm not sure this is incompatible with Windows....I opened the branch on my windows machine and I'm getting a million errors running `yarn lint`. Do you get these as well? Haven't tried on my macbook yet

image

ermish: hmmm...I'm going to test your idea for a .gitattributes file and see if this solves it

ermish: ok, I think the gitattributes won't auto fix it, but it will fix when new files are downloaded/cloned. 
I found a vscode option to use `/n` by default on windows, so we should be good! ship it 💯

ermish: we should add the `.gitattributes` in another PR like you recommended for future contributors 😸

ermish (approved):
LGTM :shipit:

PR #109, opened 2023-01-25
28 dockerize server

### LINKED ISSUE #28: Create a docker image for the graphql api server findadoc backend
- need a findadoc docker image
- then use docker-compose to run it
- ??? to the cloud???

ann-kilzer:
Resolves #28

# What changed
- [x] Add a Dockerfile to run the server
- [x] Be able to access it from outside the container
- [x] Other build optimization 
- 🏉  Make it talk to the database => Could be broken into another step

# Open questions
- do we always want this to run locally? E.g. can we make two docker-compose workflows, one for a DB only, and another for staging / prod where we want everything in docker?
-  ~GraphQL CodeGen is very chicken / egg... how do we generate types from a running server when we don't have a running server... anyway I don't quite grok this and opted to check in the generated types file for now.~
- What other things should we be doing to ensure a production run?

# Testing instructions

Build the container with:

```sh
docker build . -f ./docker/server/Dockerfile -t findadoc/server
```

Next, 

```sh
docker run -p 49150:3001 -d findadoc/server
```

This will spit out a long container id.

If you have Docker Desktop, you can go see if the container is running. Or you can check the logs like so

```sh
docker logs [container id]
```

### Goal: 

Visit http://localhost:49150 and be able to see the Apollo Explorer


src/typeDefs/gqlTypes.ts
```diff
@@ -0,0 +1,359 @@
+import { GraphQLResolveInfo } from 'graphql';
```
ann-kilzer, line 1: We either need a way we can generate this on the command line 

OR

we need to check in the generated file

Otherwise we have no way to get a reliable version of it from a container. 😤

ann-kilzer: Ohhhh @theyokohamalife and I figured it's better to point directly to the `schema.graphql` file. Using a live server is better for when you need to generate types for frontend code 💡

README.md
```diff
@@ -101,8 +101,8 @@ Read more about migrations here: https://www.prisma.io/docs/concepts/components/
 <details>
   <summary>Testing GraphQL</summary>
 
-1. Run `yarn dev` to start the local server
-2. Run `yarn generate` to generate the types locally
+1. Run `yarn generate` to generate the types locally
```
ann-kilzer, line 104: Now we can generate the types before the server runs!

docker/server/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM node:18.12.1
```
ann-kilzer, line 1: Heavily inspired by: https://nodejs.org/en/docs/guides/nodejs-docker-webapp/

package.json
```diff
@@ -7,7 +7,7 @@
         "prepare": "husky install",
         "test": "ts-mocha src/tests/**.ts",
         "dev": "yarn && ts-node-dev --transpile-only --no-notify --exit-child src/index.ts",
-        "generate": "yarn && graphql-codegen --config ./src/typesgeneratorconfig.ts",
```
ann-kilzer, line 10: Let's not call yarn inside yarn here, it tries to install husky in docker 😭

docker/server/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+# Create app directore
+WORKDIR /usr/src/app
+
+# Install app dependencies
+# A wildcard is used to ensure both package.json AND package-lock.json are copied
+# where available (npm@5+)
+COPY package*.json ./
```
RageZBla, line 9: @ann-kilzer probably worth to add `yarn.lock`

RageZBla (approved):

README.md
ermish: awesome!!

docker/server/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
```
ermish, line 9: we should actually prevent any accidental `package-lock.json` files from being added that could conflict with `yarn.lock`

Should we update to just `package.json`

docker/server/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+COPY yarn.lock ./
+
+# install ts-node
+RUN npm install -g ts-node
```
ermish, line 13: to avoid running `npm` and conflicting with yarn, we can utilize `yarn dlx` here! 
[example](https://yarnpkg.com/getting-started/migration/#use-yarn-dlx-instead-of-yarn-global)

docker/server/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+# Follow the lockfile, error on cache modify, ensure checksums match
+RUN yarn install --immutable --immutable-cache --check-cache
```
ermish, line 16: nice flags 👍

docker/server/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+# Bundle app source
+COPY . .
+
+# GraphQL types
+RUN yarn generate
+
+EXPOSE 3001
+
+CMD [ "ts-node", "src/index.ts" ]
```
ermish, line 29: fine for now 👍

package.json
ermish: haha 😅

ermish (changes requested):
awesome! nice work writing up a dockerfile 💯 

only changes needed are around the npm to yarn commands

docker/server/Dockerfile
ann-kilzer: Yeah let's improve on it going forward

docker/server/Dockerfile
ann-kilzer: Can you make a follow-up PR to get this working? I had trouble as I need to run `npm` to install yarn berry 🙃

ermish:
it's working! 
Moved the ts-node command to a yarn script
ran `docker run -p 49150:3001 -d findadoc/server`. Instead of port 8080, it should be 3001

ermish (approved):
🚀

PR #112, opened 2023-01-28
104 insurance seed data

### LINKED ISSUE #104: Use the Insurance Seed data
Great! As a follow-up, let's use the seed data from the .csv file in `devSetup.ts`. I will make a separate issue

_Originally posted by @ann-kilzer in https://github.com/ourjapanlife/findadoc-server/issues/103#issuecomment-1399478415_

Relates to #100

ann-kilzer:
Resolves #104

# What changed
- Use seed data for loading insurance
- Fix migration to remove `UNKNOWN` from enum (empty will be unknown)
- Fix an API test

# Testing instructions
1. In one terminal, start the db with `docker compose up`
2. Ensure the seeding completes 
```sh
findadoc-server-migrate-1       | {
findadoc-server-migrate-1       |   id: 2,
findadoc-server-migrate-1       |   nameJa: '博士号',
findadoc-server-migrate-1       |   nameEn: 'Doctor of Philosophy',
findadoc-server-migrate-1       |   abbreviation: 'PhD'
findadoc-server-migrate-1       | }
findadoc-server-migrate-1       |
findadoc-server-migrate-1       | 🌱  The seed command has been executed.
```

3. In another terminal, open prisma studio with `yarn prisma studio`
4. Look at the healthcare professionals and inspect the "acceptedInsurance []" and verify that the seeded data has some variable examples:

Prisma_Studio

I also fixed a test as part of this PR... perhaps I should have broken it out, but you can run it with `yarn test`


src/tests/healthcareProfessionalQuery.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
ermish, line 1: Can we fix these instead of commenting them out? 
We should either remove console logs or change the rule. 
For `any` we should either loosen the rules where they are ok in certain sections or start properly using the types. (Kinda the whole point we're using typescript right?)

src/tests/healthcareProfessionalQuery.ts
```diff
@@ -23,28 +25,32 @@ describe('GraphQL', () => {
...
-    //         // assume there are a 100 users in the database
-    //         res.body.user.should.have.lengthOf(100);
-    //       });
-    //   });
+    it('Returns all users', done => {
+        request
+            .post('/graphql')
+            .send({ query: '{ healthcareProfessionals { id } }' })
+            .expect(200)
+            .end((err:any, res: any) => {
```
ermish, line 45: We should use async/await here instead of callbacks like `end`

ermish (changes requested):
Two small changes. Check them out and see what you think

src/tests/healthcareProfessionalQuery.ts
ann-kilzer: On test files I don't care about strong typing

src/tests/healthcareProfessionalQuery.ts
ann-kilzer: Let's refactor later, this was just the example. 

We had no testing, now we have something. Feel free to open a ticket

ann-kilzer: (If you look closely at the diff you'll see I just uncommented the example and fixed it)

.eslintrc.js
```diff
@@ -119,7 +119,6 @@ module.exports = {
         'space-before-function-paren': ['error', { anonymous: 'never', named: 'never', asyncArrow: 'always' }],
         'space-in-parens': 'error',
         'space-infix-ops': 'error',
-        semi: ['error'], // personally i like them and it's safer, but can discuss
```
ann-kilzer, line 122: I'll make an issue to remove semicolons if everyone is on board

src/tests/responseTypes/responseTypes.ts
```diff
@@ -0,0 +1,8 @@
+
+export default interface responseType<Type> {
```
ann-kilzer, line 2: I wasn't quite finding what I needed in https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types, just getting lost in various `request` libs that are either deprecated or different, so I wrote a quick generic.

src/tests/healthcareProfessionalQuery.ts
ermish: that's a good question 🤔 I'm guessing there's a way to scope rules, but never looked into it.

In my personal projects, I usually make the tests `.js` files instead so I don't have to deal with types.

src/tests/healthcareProfessionalQuery.ts
ermish: ok sounds good! I'll create a separate ticket

and thanks again for writing tests ^_^

src/tests/healthcareProfessionalQuery.ts
ermish: upon a thorough 2 min investigation, I found we can do [nested configs](- I noticed we aren't `await` ing them yet. (not just the new ones), so we should probably update). So we could make a separate, more relaxed eslint config for the testing folder

ermish (approved):
Ship it!

src/tests/healthcareProfessionalQuery.ts
ann-kilzer: Oh nice! I'lll look into that, especially for the console logs, which are often useful in tests

.eslintrc.js
ann-kilzer: https://github.com/ourjapanlife/findadoc-server/issues/119

Issue #100, opened 2023-01-21
Part II of connect SQL to GraphQL


ann-kilzer:
Follow-up from #87 

# Requirements

- [x] Complete the healthcare professional #149 
- [x] Health Insurance 
- [x] Degrees => #116 
- [x] Facilities
- [x] Add PhysicalAddress to Facility
- [ ] Link Healthcare Pro and Facility? => break into a separate issue or see how it's needed by the frontend

# Dependencies

- None (Add any issue dependencies here)

Issue #78, opened 2023-01-02
Testing Framework


ann-kilzer:
- [x] Set up an API Testing framework (we have https://github.com/ladjs/supertest)
- [x] Get it running in CI/CD
- [x] Write more tests

ann-kilzer:
Looks like https://github.com/ladjs/supertest is already in here for API testing. 

Do we want unit tests? Something like Jest could be useful here and there.

And we should definitely configure it to run in GitHub Actions

Issue #46, opened 2022-12-19
Security review


ann-kilzer:
# To discuss:
- secret management
- separation of dev / prod
- access control
- PII removal requests
- repo privacy
- web security auditing
- OWASP Top 10 checks
- API parameter validation (ensure type checking and sql injection prevention)
- etc.

ermish:
added API parameter validation (ensure type checking and sql injection prevention)

ann-kilzer:
Switching from Dependabot to Mend Rennovate for security & package updates

PR #116, opened 2023-01-28
100 sql to graphql: Degrees

ann-kilzer:
Part of #100

# What changed
- Refactor Degrees to match SQL
- Add queries
- Add a test

# Testing instructions

Hey! We have a test runner in here which means #78 is already in progress 🙌🏻 

Let's start writing tests to prove our tickets are working. Then you can run with `yarn test` to verify the real graphQL query works


src/resolvers.ts
```diff
@@ -16,6 +17,16 @@ import {
 
 const resolvers = {
     Query: {
+        degrees: () => {
```
ermish, line 20: let's talk through this one. 
I'm thinking that degrees would always be queried with healthcare provider data. 

If that makes sense, 
then we should probably use your degree query within the `healthcareProfessionalService` to build the aggregated data. 

if we think degrees need to be fetched separately, 
- what would be the use cases? 
- The queries look awesome and simple. ✨

src/resolvers.ts
```diff
@@ -16,6 +17,16 @@ import {
...
+            const matchingDegrees = getDegrees();
+
+            return matchingDegrees;
+        },
+        degree: (_parent: Degree, args: { id: string; }) => {
+            const matchingDegree = getDegreeById(args.id);
```
ermish, line 26: separate ticket: I noticed we aren't `await` ing any of our queries yet. (not just these new ones), we should update them now that we've successfully connected the DB to the graph!

src/services/degreeService.ts
```diff
@@ -0,0 +1,37 @@
...
+        id: String(input.id),
+        nameJa: input.nameJa,
+        nameEn: input.nameEn,
+        abbreviation: input.abbreviation
+    } as Degree;
+}
+
+// TODO: add a validation step for incoming parameters
+export const getDegreeById = async (id: string) => {
+    const degree = await prisma.degree.findUnique(
```
ermish, line 18: future ticket: we should add in some form of data validation across all of the services so people can't sql inject.

src/tests/degreeQuery.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+
+    it('Returns degree with id = 1', done => {
+        request
+            .post('/graphql')
+            .send({
+                query: '{ degree(id: 1) { id nameJa nameEn abbreviation }}'
+            })
+            .expect(200)
+            .end((err: any, res: any) => {
```
ermish, line 23: nit: we could use `await` here instead of the `.end()` callback

src/resolvers.ts
ann-kilzer: When a user submits new data, they may need to know what degree types are available

E.g. we have some form for adding a new HealthcareProvider and we want to show the types that should be added

But maybe I'm thinking from more of a REST perspective and the InputTypes give us what we need?

Also some of these simple queries are a good teaching example for folks learning graphQL and sql, so I thought they could help us establish patterns. 🤔 but might not want to overdo it

src/resolvers.ts
ann-kilzer: I was matching the style of the current code. Do you feel `await` is better? Especially since this file is small, perhaps you could refactor it

src/services/degreeService.ts
ann-kilzer: Yes... we should also check if prisma has any built-in injection prevention-- some ORMs do

Please add to #46

src/tests/degreeQuery.ts
ann-kilzer: We could, but matching the current pattern of the tests. Let's do a later refactor

theyokohamalife:
@ann-kilzer I get an error when I attempt to run the tests. Here are the steps I've taken:

1. `yarn install`
2. `docker compose up`
3. `yarn test`

Could you let me know what step I might be missing?

Here is the error I see:

Screenshot 2023-01-29 at 9 26 46 PM

theyokohamalife:
So I realized that `yarn dev` also needs to be running. After doing that it shows 1 test failing:

Screenshot 2023-01-29 at 9 30 53 PM

src/tests/degreeQuery.ts
```diff
@@ -0,0 +1,58 @@
...
+            .send({
+                query: '{ degrees { id nameJa nameEn abbreviation }}'
+            })
+            .expect(200)
+            .end((err: any, res: any) => {
+                // res will contain array with one user
+                if (err) { return done(err); }
+
+                console.log(res.body);
+                expect(res.body.data.degrees).to.have.lengthOf(14);
```
theyokohamalife, line 51:
```suggestion
                expect(res.body.data.degrees).to.have.lengthOf(15);
```

theyokohamalife: I ran the tests and this one failed because the length of the list of degrees is actually 15. I double checked it through a query in GraphQL.

ann-kilzer: 🤔 Be sure to start your DB from scratch (delete and remake the docker container). There are only 14 entries in degrees.csv. Perhaps you added one while doing local testing

ann-kilzer: Just re-ran it on a fresh db, it's 14

theyokohamalife: @ann-kilzer  Something strange is going on, because sometimes it'll pass and sometimes it won't. I deleted the docker container between each attempt, and out of 5 tries I got:

1. Failed 
2. Passed
3. Passed
4. Failed
5. Failed

Here's my most recent failing test:

Screenshot 2023-01-29 at 10 33 37 PM

I did a little bit of digging and Prisma Studio also only shows 13 degrees in the database. Something is happening when I create a new Docker container that is sometimes causing an incorrect number of degrees from being inserted it seems.
Screenshot 2023-01-29 at 10 39 46 PM

ann-kilzer: I wonder why `Registered Nurse` isn't showing up in your DB. Everything from `degrees.csv` should be loaded in the db...

ann-kilzer: Possibly a race condition in forEach...

ann-kilzer: I refactored the seed data function. I think `forEach` with the implied sql ids meant that there was a race condition.

Try it again on a fresh DB, I think it's working. Will have to refactor the seed data for the other methods later

theyokohamalife (approved):
It works now. Thanks! 😎

PR #126, opened 2023-01-31
chore: replace apollo-server with @apollo/server

### LINKED ISSUE #107: Migrate off Apollo Deprecated Server Plugins
I noticed some warnings on yarn install

```
=> # warning apollo-server > apollo-server-core > apollo-server-plugin-base@3.7.1: The `apollo-server-plugin-base` package is part of Ap
 => => # ollo Server v2 and v3, which are now deprecated (end-of-life October 22nd 2023). This package's functionality is now found in the `
 => => # @apollo/server` package. See https://www.apollographql.com/docs/apollo-server/previous-versions/ for more details.
 => => # warning apollo-server > apollo-server-core > apollo-server-plugin-base > apollo-server-types@3.7.1: The `apollo-s
 ```

# Requirements

- [ ] Update libraries to use the ones that will be supported beyond 2023

Please Check `yarn install` for all errors related to Apollo 

# Dependencies

### PR DESCRIPTION
**Pull request by theyokohamalife**
Resolves #107 

# What changed
1. [apollo-server](https://www.npmjs.com/package/apollo-server) has been deprecated and replaced with [@apollo/server](https://www.npmjs.com/package/@apollo/server).
2. The [Get Started](https://www.apollographql.com/docs/apollo-server/getting-started) documentation for @apollo/server recommends using `await startStandaloneServer` to initialize the server, but some changes will need to be done to allow top-level `await` in our repo. For now, I settled for wrapping the await in an async function, but this should probably be refactored by someone more familiar with Typescript.
3. The **graphql-tag** package has been added, because `gql` is not part of @apollo/server.

# Testing instructions

1. Delete `node_modules`
2. Run `yarn install`
4. Run `yarn dev`
5. Confirm that GraphQL Studio is working

### CODE REVIEW

**Comment by theyokohamalife** [comment]
Not sure if I should install the package recommended by Socket Security or not.
Screenshot 2023-02-01 at 2 26 32 AM

**Comment by ann-kilzer** [comment]
I don't think we need protocol buffers in a small project like this

**Comment by theyokohamalife** [comment]
> I don't think we need protocol buffers in a small project like this

@ann-kilzer Could you elaborate? I'm not sure what's considered a protocol buffer.

**Comment by ann-kilzer** [comment]
Read more about them here: https://developers.google.com/protocol-buffers

The tldr is it's a way of serializing data that's good for large-scale software. Let's stick with JSON for mvp

**Comment by theyokohamalife** [comment]
@SocketSecurity ignore @apollo/protobufjs@1.2.7

File: src/index.ts [hunk 1]
```diff
@@ -10,7 +11,14 @@ const server = new ApolloServer({
 
 const port = 3001
 
-server.listen({ port }).then(({ url }) => {
-    // eslint-disable-next-line no-console
-    console.log(`🚀  Server ready at ${url}`)
-})
+async function getUrl() {
```
ann-kilzer, line 14: As a general rule, "getters" should not have side effects. Here, we have a very big side effect: starting a web server, and then we're not even using the `url` we got.

At minimum I suggest a rename

```suggestion
async function startServer() {
```

theyokohamalife: @ann-kilzer  Could you take another look at it? I refactored it with your suggestions.

src/index.ts
```diff
@@ -10,7 +11,12 @@ const server = new ApolloServer({
...
-})
+function startServer() {
+    return startStandaloneServer(server, {
+        listen: { port: port }
+    })
+}
+
+// eslint-disable-next-line no-console
+startServer().then(() => console.log(`🚀  Server ready at: http://localhost:${port}`))
+
```
ann-kilzer, line 22: Better! Here's another alternative using await. Let me know which you think looks better, they both work.

```suggestion
async function startServer(port = 3001) {
    await startStandaloneServer(server, {
        listen: { port: port }
    })

    // eslint-disable-next-line no-console
    console.log(`🚀  Server ready at: http://localhost:${port}`)
}

startServer()
```

ann-kilzer: Also what do you think about renaming this file `main.ts`? `index` has such strong associations with frontend HTML `/` routes.

ann-kilzer (approved):
Approving with some optional comments

src/index.ts
ermish: this looks good to me.  I noticed you didn't await `startServer()` which I think is the right way to do it 👍

ermish: index is also pretty common in nodejs apps for reference. I don't have a preference either way, just as long as it's easy to understand what is the entry point file

ermish (approved):
also LGTM. I like Ann's suggestion if we can do it in this PR or the next so we aren't using callbacks

src/index.ts
ann-kilzer: We can't await at the top level in our version of ecmascript

PR #129, opened 2023-02-02
Docker db connection

### LINKED ISSUE #128: Get docker server to communicate with docker postgres
Follow-up from #109 

# Requirements

- [x] docker server 🤝 docker postgres

# Dependencies

ann-kilzer:
Resolves #128

# What changed
- [x] Connect docker server to docker db

# Testing instructions

1. In a terminal, run `docker compose up` to start the database
2. Copy your postgres password from your `.env` file
3. Run the build command, replacing `[your password]` with the password
```sh
yarn docker:build --build-arg pgPassword=[your password]
```
4. Start the container:
```sh
yarn docker:run
```
5. Open up http://localhost:49120 and run a query in the apollo server

Screenshot 2023-02-04 at 11 12 03 PM


docker/server/Dockerfile
```diff
@@ -1,5 +1,15 @@
 FROM node:18.12.1
 
+# ARGS
+ARG pgUser=postgres
```
ermish, line 4: awesome! this is how I like to set it up as well :) nice work

docker/server/Dockerfile
```diff
@@ -1,5 +1,15 @@
...
+ARG pgPassword
+ARG pgHost=database
```
ermish, line 6: future improvement: to make it easier, I usually like to set the args default from the .env file, then the args are just optional overrides

ermish:
>  `In a terminal, run docker compose up to start the database` 
future issue: we should put the db startup into the yarn scripts as well....
I imagine there being something like  a `docker` script for everything in compose, `docker:serveronly` and `docker:dbonly`.....

ermish:
It connected!!! 💯 

I didn't get the proper query schema though. 🤔 

image

ermish (approved):
good iteration!

ann-kilzer:
> > `In a terminal, run docker compose up to start the database`
> 
> future issue: we should put the db startup into the yarn scripts as well.... I imagine there being something like a `docker` script for everything in compose, `docker:serveronly` and `docker:dbonly`.....

Agree, just wasn't quite sure how to get `.env` vars into the package.json without hard-coding secrets... unless we make another docker-compose, and that is additional work

docker/server/Dockerfile
ann-kilzer: Can you explain how you do that?

PR #170, opened 2023-04-08
Seed firestore database

jbcden:
# What changed

This currently does not resolve any open issues but is helping work towards getting database seeding ready for firestore. In addition, this is adding some basic queries for the seeded data.

Currently there are some limitations with the current implementation. The current queries need to be updated to handle return the document id so that mutations can be applied (I think). In addition, the seeding is clumsy as it is difficult to associate resources together and running the seeding script would probably result in creating duplicate documents (there's probably a way around that but I'm not sure).

# Testing instructions


src/database.ts
```diff
@@ -0,0 +1,88 @@
...
+  return healthcareProfessionals
+}
+
+export async function getFacilities() {
+    const db = getFirestore()
+    const facilitiesRef = db.collection('facilities')
+
+    const snapshot = await facilitiesRef.get()
+    var facilities = []
+    snapshot.forEach(doc => {
```
ermish, line 45: Js optimization 
```suggestion
    const facilities = snapshot.map(doc => { ...}
```

src/database.ts
```diff
@@ -0,0 +1,88 @@
...
+      facilities.push(doc.data())
+    })
+
+    facilities = Promise.all(facilities.map(facility => hydrateFacility(facility))).then((resource) => resource)
```
ermish, line 49: Nice catch! This is the proper wait o await a foreach loop. BUT! They just released a new syntax to solve this ✨

```suggestion
    for await (const facility of facilities) {
    await yourcode ()
  }
```

src/database.ts
```diff
@@ -0,0 +1,88 @@
...
+    // console.log(facilities)
+
+    return facilities
+}
+
+async function hydrateFacility(facility) {
+    const db = getFirestore()
+    const hpRef = db.collection('healthcareProfessionals')
+    const snapshot = await hpRef.where('id', 'in', facility.healthcareProfessionals).get()
```
ermish, line 58: Nice! Looks like you figured out the queries ね

src/database.ts
```diff
@@ -0,0 +1,88 @@
...
+    const snapshot = await hpRef.get()
+  var healthcareProfessionals = []
+    snapshot.forEach(doc => {
+      healthcareProfessionals.push(doc.data())
+    })
+
+  return healthcareProfessionals
+}
+
+export async function getFacilities() {
```
ermish, line 39: For building queries/commands, we can move these to the respective domain services. `Facilityservice.ts`, etc. 

This allows the `database.ts` to have a single responsibility of managing db connections.

src/fakeData/facilities.ts
```diff
@@ -0,0 +1,58 @@
+import { HealthcareProfessional, Locale, LocaleName, Degree, 
+    Specialty, SpecialtyName, SpokenLanguage, Insurance, Facility } from '../typeDefs/dbSchema'
+
+export const fakeFacilities = () => {
+    // const name : LocaleName = {
```
ermish, line 5: You could delete these comments since we have a different method to create a healthcare prof

ermish (approved):
Looks good so far! Feel free to merge and fix in follow up this time since we're not breaking anything if it's not running 😎

src/database.ts
jbcden: Surprisingly, while `snapshot.forEach` is defined here, `snapshot.map` is undefined.

PR #203, opened 2023-08-02
feat: add mutations for addFacility and addHealthcareProfessional

### LINKED ISSUE #207: Write a mutation for AddFacility

### LINKED ISSUE #211: Write mutation for addHealthcareProfessional

theyokohamalife:
Resolves #207 
Resolves #211 

# What changed
1. Creates mutation for addFacility that will save a new Facility to Firestore with a list of HealthcareProfessional IDs.
2. Creates mutation for addHealthcareProfessional that will save a new HealthcareProfessional to Firestore.

# Testing instructions
1. Paste the following into the `Variables` section at the bottom of GraphQL studio to create a new Facility and HealthcareProfessional. 
2. Check Firestore and verify that the new documents were created.

```
{
  "input": {
    "nameJa": "some facility ja name",
    "nameEn": "some facility en name",
    "contact": {
      "website": "http://some-website",
      "phone": "000-0000-0000",
      "mapsLink": "http://some-map",
      "email": "some@email.address",
      "address": {
        "addressLine1En": "123 my address",
        "addressLine1Ja": "123アドレス",
        "addressLine2En": "Amazing House",
        "addressLine2Ja": "アメージングハウス",
        "cityEn": "some city en",
        "cityJa": "some city ja",
        "postalCode": "12345",
        "prefectureEn": "some prefecture en",
        "prefectureJa": "some prefecture ja"
      }
    },
    "healthcareProfessionals": [
      {
        "acceptedInsurance": ["INSURANCE_NOT_ACCEPTED", "INTERNATIONAL_HEALTH_INSURANCE"],
        "degrees": [
          {
            "nameJa": "some degree ja",
            "nameEn": "some degree en",
            "id": "1",
            "abbreviation": "S.D."
          }
        ],
        "names": [
          {
            "middleName": "some middle name",
            "locale": "en",
            "lastName": "some last name",
            "firstName": "some first name"
          }
        ],
        "spokenLanguages": [
          {
            "iso639_3": "en",
            "nameEn": "English",
            "nameJa": "English in ja",
            "nameNative": "English"
          }
        ],
        "specialties": [
          {
            "names": [
              {
                "name": "best doctor in the world",
                "locale": "some locale"
              }
            ]
          }
        ]
      }
    ]
  }
}
```


src/resolvers.ts
```diff
@@ -1,114 +1,76 @@
...
     Facility,
     Insurance,
     HealthcareProfessional,
     Specialty,
     LocaleNameInput,
     HealthcareProfessionalInput,
-    SpokenLanguage
+    SpokenLanguage,
+    LocaleName,
+    DegreeInput,
```
ermish, line 16: what's your thoughts on why we created separate input entities vs exposing the regular entities?

theyokohamalife: @ermish I'm not sure what you mean. I thought we had to create the input entities? I just followed the [doc examples](https://graphql.org/graphql-js/mutations-and-input-types/). Are you asking this because they're unnecessary?

ermish: it depends on your design, but there's 3 options:
1 -  have separate entities for your api, business logic, and database
2 -  share entities for your business logic and db
3 -  share entities for your api and business logic

I usually prefer #3 or #1. If we want them separate for input that's fine. We should do this is across our api then for any inputs to be clearly suffixed with `Input`

src/services/facilityService.ts
```diff
@@ -18,14 +19,31 @@ export const getFacilityById = async (id: string) : Promise<Facility | null> =>
...
 
-export const addFacility = async (facility : Facility) : Promise<void> => {
-    //todo
+export const addFacility = async (input: FacilityInput) : Promise<Facility> => {
+    const db = getFirestore()
+   
+    const facilityRef = db.collection('facilities')
+
+    const healthcareProfessionalIds = 
+    await mapAndValidateHealthcareProInput(input.healthcareProfessionals as HealthcareProfessionalInput[])
```
ermish, line 28: nit: should stay consistent with either professsional or pro

src/services/facilityService.ts
```diff
@@ -41,12 +59,41 @@ export const searchFacilities = async (userSearchQuery : string[]) : Promise<Fac
...
+    return nameEn
+}
+
+function validateNameJaInput(nameJaInput: string) : string {
+    const nameJa = nameJaInput
+
+    return nameJa
+}
+
+function mapAndValidateHealthcareProInput(healthcareProInput: HealthcareProfessionalInput[]) : Promise<string[]> {
```
ermish, line 95: since we have a healthcare professional service, we should move this there, and call it from here. (also the dependent sub methods as well)

src/services/healthcareProfessionalService.ts
```diff
@@ -17,9 +23,24 @@ export const getHealthcareProfessionalById = async (id: string) : Promise<Health
...
+        acceptedInsurance: mapAndValidateInsurance(input.acceptedInsurance as []),
+        degrees: mapAndValidateDegrees(input.degrees as DegreeInput[]),
+        names: mapAndValidateNames(input.names as LocaleNameInput[]),
+        specialties: mapAndValidateSpecialties(input.specialties as SpecialtyInput[]),
+        spokenLanguages: mapAndValidateLanguages(input.spokenLanguages as SpokenLanguageInput[])
+    }
+
+    const idList: string[] = []
+    
+    await healthcareProRef.add(newHealthcareProfessional).then((docRef: DocumentReference) => idList.push(docRef.id))
```
ermish, line 41: don't use `.then()` with `await`. They're conflicting designs. Try to use async/await where possible

src/typeDefs/dbSchema.ts
```diff
@@ -1,20 +1,21 @@
...
+  address: PhysicalAddress
 }
 
 export type Facility = {
-  id: string
   nameEn: string
   nameJa: string
   contact: Contact
-  healthcareProfessionals: HealthcareProfessional[]
+  healthcareProfessionalIds: string[]
```
ermish, line 15: we likely want ids AND professsionals. When you get a facility, we'll want to have the entities and not just IDs. The alternative would be to use one list and just populate the rest of the data besides the id later, but I think this confuses most people. The other alternative is to make an `Output` entity for each API get request, but I don't think we need to do that now. 

My vote is to have the entity here match what is returned to the user

src/services/facilityService.ts
theyokohamalife: The reason for the change was because eslint was complaining that it was too long 😅

src/resolvers.ts
theyokohamalife: @ermish You're pretty much speaking a foreign language to me. 😵 Is it okay to keep the `Input`? They've existed in the schema for some time now for each Type. Do you see anywhere that the suffix is missing?

src/typeDefs/dbSchema.ts
theyokohamalife: @ermish Hmm I'm pretty sure that at our last dev meeting you said `Facility` should have a list of `HealthcareProfessional` IDs. When we looked through Firestore together I showed you that the HealthcareProfessional was saved to both the `facilities` and `healthcareProfessionals` collections. So just to be clear, you want all of the `HealthcareProfessional` data saved to both collections?

src/resolvers.ts
ermish: TLDR, any input coming from the user should have the suffix `Input`. I think you have them covered here so far

src/typeDefs/dbSchema.ts
ermish: no, sorry. 
two arrays

1) `healthcareProfessionalIds` gets saved to the db in the facility. 
2) `healthcareProfessionals` is the array that gets populated when facilities are fetched.  (or coming from the user input)

theyokohamalife: @ermish So, just to make sure we're on the same page, `healthcareProfessionalIds` is already being saved to the db in the facility. The only change you want is to fetch the professionals by ID when a facility is queried, right? You don't want the healthcareProfessional entities to be duplicated across collections, right?

So the schema changes you want is this?

```
type Facility
healthcareProfessionalIds: string[]
healthcareProfessionals: HealthcareProfessional[]
```

ermish: thats right

ermish (approved):
:shipit:

Issue #212, opened 2023-08-08
Create a service file for Submission


theyokohamalife:
- [x] getSubmission
- [x] getSubmissions
- [x] addSubmission
- [x] updateSubmission
- [x] mapDbEntityToGQLEntity (ask Philip for more info)

PR #213, opened 2023-08-08
feat: add a submissionService file and Submission type definitions

evan-desu:
Previously there were no service functions to allow users to get, add, or update submissions. There are now functions to perform these tasks. There is now also a Submission type alias in `TypeDefs/dbschema.ts`.

Part of #212 

This PR is part of a 🥞:
1. https://github.com/ourjapanlife/findadoc-server/pull/213 <- You are here
2. https://github.com/ourjapanlife/findadoc-server/pull/218 

# What changed
1. Added a type alias called `Submission` to `dbSchema.ts` to define the data types for each field.
3. Created a file in `src/services` called `submissionService.ts` to house the submission-related functions.
4. Added three main functions: 
- **`getSubmissionById`**: Retrieves a specific submission by its ID.
- **`addSubmission`**: Adds a new submission.
- **`updateSubmission`** : Updates an existing submission.
5. Added the `mapDbEntityTogqlEntity` function to ensure that data is structured in the desired format for GraphQL presentation.
6. Added `submission`, `createSubmission`, and `updateSubmission` to `resolvers.ts`.

# Testing instructions
1. Run the development server using `yarn dev`.
2. Open your browser and navigate to http://localhost:3001 to access the GraphQL Studio.
3. In the GraphQL Studio, click the explorer button on the left.
4. Under **Root Types**, click the `+` next to `mutation: Mutation`.
5. Under **Fields**, click the `+` next to `createSubmission`.
7. in the middle pane, under **Operation**, copy and past the following:
```
mutation Mutation($input: SubmissionInput) {
    id
    googleMapsUrl
    healthcareProfessionalName
    spokenLanguages {
      iso639_3
      nameJa
      nameEn
      nameNative
    }
    isUnderReview
    isApproved
    isRejected
  }
}
```
8. In the middle pane, under **Variables**, copy and paste the following:
```
{
  "input": {
    "googleMapsUrl": "https://maps.google.com/example",
    "healthcareProfessionalName": "Enter Name Here",
    "spokenLanguages": [
      {
        "iso639_3": "eng",
        "nameJa": "English",
        "nameEn": "English",
        "nameNative": "English"
      }
    ]
  }
}
```
9. Click the blue **Mutation** button. The result in the right pane should match the structure of this example:
```
{
  "data": {
    "createSubmission": {
      "id": "actual_id_here",
      "googleMapsUrl": "https://maps.google.com/kitty-clinic",
      "healthcareProfessionalName": "Dr. Carlos",
      "spokenLanguages": [
        {
          "iso639_3": "eng",
          "nameJa": "meow meow meow",
          "nameEn": "meow meow meow",
          "nameNative": "meow meow meow"
        }
      ],
      "isUnderReview": true,
      "isApproved": false,
      "isRejected": false
    }
  }
}
```
10. Check the **submissions collection** in Firestore to confirm successful creation.

### DIFF
--- a/src/resolvers.ts
+++ b/src/resolvers.ts
@@ -1,6 +1,7 @@
 import * as facilityService from './services/facilityService'
 import * as healthcareProfessionalService from './services/healthcareProfessionalService'
 import * as gqlType from './typeDefs/gqlTypes'
+import * as submissionService from './services/submissionService'
 
 const resolvers = {
     Query: {
@@ -24,6 +25,11 @@ const resolvers = {
             await healthcareProfessionalService.getHealthcareProfessionalById(args.id)
 
             return matchingHealthcareProfessional
+        },
+        submission: async (_parent: gqlType.Submission, args: { id: string }) => {
+            const matchingSubmission = await submissionService.getSubmissionById(args.id)
+            
+            return matchingSubmission
         }
     },
     Mutation: {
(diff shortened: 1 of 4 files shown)


src/services/submissionService.ts
```diff
@@ -0,0 +1,62 @@
...
+    })
+
+    return submissions
+}
+
+export const addSubmission = async (submission: Submission) : Promise<void> => {
+    const db = getFirestore()
+    const submissionRef = db.collection('submissions')
+
+    await submissionRef.add(submission)
```
ermish, line 40: we should actually send back a response for the client to know if it was successful. 
I can do this later 👍

src/typeDefs/schema.graphql
```diff
@@ -92,6 +92,22 @@ type SpokenLanguage {
   nameNative: String!
 }
 
+type Submission {
+  id: ID!
+  submittedLink: String!
```
ermish, line 97: what kind of link? This should be something like `googleMapsUrl`

src/services/submissionService.ts
evan-desu: Thanks for the feedback, Philip! I can add this tomorrow. I'm thinking a try/catch block would be good as it would include error handling.

src/typeDefs/schema.graphql
evan-desu: Yes, that's definitely more descriptive. I'll make the change as soon as I have the chance.

src/resolvers.ts
```diff
@@ -41,6 +44,16 @@ const resolvers = {
             const matchingHealthcareProfessional = await getHealthcareProfessionalById(args.id)
 
             return matchingHealthcareProfessional
+        },
+        // submissions: async () => {
```
ermish, line 48: let's delete this commented out code

src/resolvers.ts
```diff
@@ -69,9 +82,75 @@ const resolvers = {
...
+                healthcareProfessionalName?: string,
+                spokenLanguages?: SpokenLanguageInput[],
+                isUnderReview?: boolean,
+                isApproved?: boolean,
+                isRejected?: boolean
+            }
+        }) => {
+            let updatedSpokenLanguages
+
+            if (args.input.spokenLanguages) {
```
ermish, line 127: good thinking here!
you can move this validation logic into the submission service as a `mapAndValidateSpokenLanguages` method and call it from  here. 
Also, consider what happens if it fails, what should we return? likely we'll need to send some sort of error to the user and not run the update

src/services/submissionService.ts
```diff
@@ -0,0 +1,85 @@
...
+        const updatedSubmission: Submission = {
+            ...submissionToUpdate,
+            ...updatedFields
+        }
+
+        await submissionRef.update(updatedSubmission)
+
+        return 'Submission updated successfully!'
+    } catch (error) {
+        throw new Error(`Failed to update submission: ${error}.`)
```
ermish, line 69: you'll want to wrap the call to this method (in the resolver) with a try/catch so it doesn't shut down the API when there's an error. 
This is a good practice in general and we'll probably want all of our resolvers to follow this pattern. (the alternative design instead of throwing an Error would be that every method returns a "result" style object with a  pass/fail boolean to continue.) they both have tradeoffs. Let's start with try/catch for now since it's easier.

ermish (changes requested):
almost there! just a few things to cleanup

src/typeDefs/dbSchema.ts
```diff
@@ -54,6 +54,22 @@ export type SpokenLanguage = {
...
+  id: string,
+  googleMapsUrl: string,
+  healthcareProfessionalName: string,
+  spokenLanguages: SpokenLanguage[],
+  isUnderReview: boolean,
+  isApproved: boolean,
+  isRejected: boolean
+}
+
+export type SubmissionFilters = {
```
ermish, line 67: Is this being used?

evan-desu:
SubmissionFilters is not being used here, but is in the next PR.
________________________________

PR #223, opened 2023-08-19
Refactor FacilityService and HealthcareProfessionalService

### LINKED ISSUE #219: Refactor addFacility functionality
# Requirements

- [ ] Refactor the `addFacility` method in the `facilityService`
    - [ ] Remove saving degrees, language, specialties, and any associated data from the facility and healthcare professional
    - [ ] To save a facility
        - [ ] first save the healthcare professionals to the `healthcareProfessional` db collection
        - [ ] get the ids of the saved professionals and add to the facility `healthcareProfessionalIds` list. 
        - [ ] Save the facility with the new healthcareProfessional ids. NOTE: the healthcare professionals list should NOT be saved in the facility collection. This is used for querying.

theyokohamalife:
Resolves #219

# What changed
1. `addFacility` now saves a healthcareProfessional to the `healthcareProfessionals` database collection, and then stores the id in an array to the new Facility in the `facilities` collection.
2. `addHealthcareProfessionalToFacility` was created as a separate function from `addHealthcareProfessional` to allow a HealthcareProfessional to be added to an existing Facility by passing the `facilityId`.

# Testing instructions
Operation: 
```
mutation CreateFacilityWithHealthcareProfessional($input: FacilityInput) {
  createFacilityWithHealthcareProfessional(input: $input) {
    id
  }
}
```

Variables:
```
{
  "input": {
    "nameJa": "some facility name",
    "nameEn": "some facility name",
    "contact": {
      "email": "email",
      "address": {
        "addressLine1En": "some address",
        "addressLine1Ja": "some address",
        "addressLine2En": "some address",
        "addressLine2Ja": "some address",
        "cityEn": "some address",
        "cityJa": "some address",
        "postalCode": "some address",
        "prefectureEn": "some address",
        "prefectureJa": "some address"
      },
      "mapsLink": "some contact",
      "phone": "some contact",
      "website": "some contact"
    },
    "healthcareProfessionals": [
      {
        "acceptedInsurance": "INSURANCE_NOT_ACCEPTED",
        "degrees": [
          {
            "nameJa": "some degree",
            "nameEn": "some degree",
            "abbreviation": "some degree"
          }
        ],
        "spokenLanguages": [
          {
            "nameNative": "some language",
            "nameJa": "some language",
            "nameEn": "some language",
            "iso639_3": "some language"
          }
        ],
        "specialties": [
          {
            "names": [
              {
                "name": "some specialty",
                "locale": "ENGLISH"
              }
            ]
          }
        ],
        "names": [
          {
            "middleName": "some person name",
            "locale": "ENGLISH",
            "lastName": "some person name",
            "firstName": "some person name"
          }
        ]
      }
    ],
    "isDeleted": false
  }
}
```

### DIFF
--- a/src/resolvers.ts
+++ b/src/resolvers.ts
@@ -1,24 +1,8 @@
 import { addFacility, getFacilityById, searchFacilities } from './services/facilityService'
-import { addHealthcareProfessional, 
-    getHealthcareProfessionalById,
+import { addHealthcareProfessionalToFacility } from './services/healthcareProfessionalService'
+import {getHealthcareProfessionalById,
     searchHealthcareProfessionals } from './services/healthcareProfessionalService'
-import {
-    Contact,
-    Degree,
-    Facility,
-    Insurance,
-    HealthcareProfessional,
-    Specialty,
-    LocaleNameInput,
-    HealthcareProfessionalInput,
-    SpokenLanguage,
-    LocaleName,
-    DegreeInput,
-    SpecialtyInput,
-    SpokenLanguageInput,
-    FacilityInput,
-    ContactInput
-} from './typeDefs/gqlTypes'
+import * as gqlType from './typeDefs/gqlTypes'
 
 const resolvers = {
     Query: {
@@ -27,7 +11,7 @@ const resolvers = {
 
             return matchingFacilities
         },
-        facility: async (_parent: Facility, args: { id: string; }) => {
+        facility: async (_parent: gqlType.Facility, args: { id: string; }) => {
             const matchingFacility = await getFacilityById(args.id)
 
             return matchingFacility
@@ -37,17 +21,17 @@ const resolvers = {
 
             return matchingProfessionals
         },
-        healthcareProfessional: async (_parent: HealthcareProfessional, args: { id: string; }) => {
+        healthcareProfessional: async (_parent: gqlType.HealthcareProfessional, args: { id: string; }) => {
             const matchingHealthcareProfessional = await getHealthcareProfessionalById(args.id)
 
             return matchingHealthcareProfessional
         }
     },
     Mutation: {
-        createFacility: async (_parent: FacilityInput, args: {
+        createFacilityWithHealthcareProfessional: async (_parent: gqlType.Facility, args: {
             input: {
-                contact: ContactInput,
-                healthcareProfessionals: HealthcareProfessionalInput[],
+                contact: gqlType.Contact,
+                healthcareProfessionals: gqlType.HealthcareProfessional[],
                 nameEn: string,
                 nameJa: string,
             }
@@ -56,17 +40,18 @@ const resolvers = {
 
             return newFacility
         },
-        createHealthcareProfessional: async (_parent: HealthcareProfessionalInput, args: {
+        createHealthcareProfessional: async (_parent: gqlType.HealthcareProfessional, args: {
             input:{
-                acceptedInsurance: Insurance[],
-                degrees: Degree[],
-                names: LocaleName[]
-                specialties: SpecialtyInput[]
-                spokenLanguages: SpokenLanguageInput[]
+                facilityId: string,
+                acceptedInsurance: gqlType.Insurance[],
+                degrees: gqlType.Degree[],
+                names: gqlType.LocaleName[]
+                specialties: gqlType.Specialty[]
+                spokenLanguages: gqlType.SpokenLanguage[]
 
             }
         }) => {
-            const newHealthcareProfessional = await addHealthcareProfessional(args.input)
+            const newHealthcareProfessional = await addHealthcareProfessionalToFacility(args.input)
 
             return newHealthcareProfessional
         }
--- a/src/resolvers.ts
+++ b/src/resolvers.ts
@@ -1,28 +1,26 @@
-import { addFacility, getFacilityById, searchFacilities } from './services/facilityService'
-import { addHealthcareProfessionalToFacility } from './services/healthcareProfessionalService'
-import {getHealthcareProfessionalById,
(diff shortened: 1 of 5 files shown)


src/resolvers.ts
```diff
@@ -1,72 +1,56 @@
...
-    SpokenLanguage,
-    LocaleName,
-    DegreeInput,
-    SpecialtyInput,
-    SpokenLanguageInput,
-    FacilityInput,
-    ContactInput
-} from './typeDefs/gqlTypes'
+import * as facility from './services/facilityService'
+import * as healthcareProfessional from './services/healthcareProfessionalService'
```
ermish, line 2:
```suggestion
import * as healthcareProfessionalService from './services/healthcareProfessionalService'
```

src/resolvers.ts
```diff
@@ -1,72 +1,56 @@
...
-    HealthcareProfessionalInput,
-    SpokenLanguage,
-    LocaleName,
-    DegreeInput,
-    SpecialtyInput,
-    SpokenLanguageInput,
-    FacilityInput,
-    ContactInput
-} from './typeDefs/gqlTypes'
+import * as facility from './services/facilityService'
```
ermish, line 1: Should make it clear that these are from a service. facility just sounds like an object/entity.

```suggestion
import * as facilityService from './services/facilityService'
```

src/services/facilityService.ts
```diff
@@ -19,36 +17,42 @@ export const getFacilityById = async (id: string) : Promise<Facility | null> =>
...
+export async function addFacility(input: gqlTypes.Facility) {
+    const db = firebase.getFirestore()
    
-    const facilityRef = db.collection('facilities')
-
-    const healthcareProfessionalIds = await mapAndValidateHealthcareProInput(
-        input.healthcareProfessionals as HealthcareProfessionalInput[]
-    )
+    const facilityRef = db.collection('facilities').doc()
+    const healthcareProfessionalRef = db.collection('healthcareProfessionals').doc()
```
ermish, line 24: should move this logic into the healthcareProfessionalService. 

Likely should make a method `addHealthcareProfessionals` that can take an array of them, validate them, and save them.

src/services/facilityService.ts
```diff
@@ -19,36 +17,42 @@ export const getFacilityById = async (id: string) : Promise<Facility | null> =>
...
+
+    if (input.healthcareProfessionals !== null 
+        && input.healthcareProfessionals !== undefined 
+        && input.healthcareProfessionals.length > 0) {
+        addHealthcareProfessional(
```
ermish, line 29: we'll want to be able to add more than one. We should do this for the whole array not just the first item. 

you'll probably need an awaited for loop for each item to save...like this
https://stackoverflow.com/a/50874507/3417143

src/typeDefs/schema.graphql
```diff
@@ -46,22 +46,26 @@ type Facility {
   nameJa: String
   contact: Contact
   healthcareProfessionalIds: [String]
-  healthcareProfessionals: [HealthcareProfessional]
+  healthcareProfessionals: [HealthcareProfessional],
+  isDeleted: Boolean
```
ermish, line 50: I think we decided against using the `isDeleted` flag 😅  we're going to use logging to track deleted items

src/services/healthcareProfessionalService.ts
```diff
@@ -1,18 +1,11 @@
...
-    const whereCondition = '=' as WhereFilterOp
-    const snapshot = await hpRef.where('id', whereCondition, id).get()
+import * as firebase from 'firebase-admin/firestore'
+import * as typeDefs from '../typeDefs/gqlTypes'
+
+export async function getHealthcareProfessionalById(id: string) {
+    const db = firebase.getFirestore()
+    const healthcareProfessionalRef = db.collection('healthcareProfessionals')
+    const whereCondition = '=' as firebase.WhereFilterOp
+    const snapshot = await healthcareProfessionalRef.where('id', whereCondition, id).get()
```
ermish, line 8: nit: the "conditions" setup isn't consistent across the different methods

ermish (changes requested):
nice work!

src/services/facilityService.ts
theyokohamalife: @ermish I tried that initially and couldn't get it to return the healthcareProfessional Ids instead of a promise even with async/await.

src/services/facilityService.ts
theyokohamalife: @ermish See my previous comment. I couldn't get it working because it would only return a promise. I'd like to keep it as one for now and if you could please refactor it when you jump into the backend work. 🙏🏾

src/typeDefs/schema.graphql
theyokohamalife: @ermish I don't think we clearly decided against the `isDeleted` flag? I think we just agreed to remove it from the Facility list. However, doing that alone wouldn't prevent the entry from appearing in the HealthcareProfessional query.

Screenshot 2023-08-21 at 3 27 46 PM

How will the logger know if a HealthcareProfessional is deleted if there isn't a property anywhere that says it is?

src/services/facilityService.ts
ermish: Sounds good 👍 I can update in a separate PR

src/typeDefs/schema.graphql
ermish: You would have a log of the id being removed from a facility and then a log of the deleted healthcareprofessional entity

ermish (approved):
Let's get it shipped 👍

src/typeDefs/schema.graphql
theyokohamalife: @ermish Okay, I'll open a new issue that includes removing this field when the logger is added.

PR #228, opened 2023-08-21
Remove codeowners until the repo goes public

theyokohamalife:
# What changed

Temporarily removing CODEOWNERS, because it's placing a burden on development with a team so small on a private repository.


ermish (approved):

ermish:
as an alternative, I think we can just change the rules to not require a review instead. However, I think we should still merge this because I think using GitHub teams for different permissions is the newer/better way to maintain rather than a code owners file.

PR #248, opened 2023-09-04
docs: update instructions for emulator

### LINKED ISSUE #246: Update: README firestore command when not seeding

Anissa3005:
Prior to this commit some devs were experiencing problems with the seeds not appearing in the emulator. This adds instructions to the README to solve this problem.

Resolves #246

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,12 +114,6 @@ yarn dev
 
 This will run until you shut down the instance hitting `ctrl^+C`
 
-If the emulator is missing the seed data try to run emulator with the following added to make sure the correct project id is set:
-
-```sh
-yarn dev:startlocaldb --project="find-a-doc-japan"
-```
-
 ### (Optional) Connecting to Production database
 
 #### Setting Up Firebase Service Account
--- a/README.md
+++ b/README.md
@@ -35,10 +35,10 @@ We love and welcome contributions to our front-end repository which can be found
 
 ## Prerequisites
 
-- [Node](https://nodejs.org)
-  - We recommend using [nvm](https://github.com/nvm-sh/nvm) and running `nvm use` in this directory to sync with the project's Node version. However, if you wish to install Node your own way and ensure a consistent version with `.nvmrc`, that's fine too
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -102,6 +105,18 @@ Select "Use an existing project"
...
+### 3.c Starting the dev server and seeding the emulator
+
+```sh
+yarn dev
+```
+
+If the emulator is missing the seed data try to run the following to make sure the correct project id is set:
+
+```sh
+yarn firebase --project="find-a-doc-japan" emulators:start
```
ermish, line 117: is this command correct? I'm not sure if yarn will run this if it's not a script. Please double check. If it works, this is fine 👍

ermish: actually, as I'm thinking about it, can we just add the project option to the `yarn dev:startlocaldb` command itself?

theyokohamalife: @ermish I think it's a good idea to add it directly to the script!

Anissa3005: @ermish @theyokohamalife I double checked and the command is working. But I feel I should make it more clear when to run that command? 
Since you have to run `yarn firebase --project="find-a-doc-japan" emulators:start ` instead of `yarn dev:startlocaldb`  followed by `yarn dev`

ermish: @Anissa3005 the idea is you can add the `--project="find-a-doc-japan"` to the `dev:startlocaldb` so you won't need to run your firebase command anymore 😄  then we can just run `dev:startlocaldb` then `yarn dev`

Anissa3005: @ermish I see since `dev:startlocaldb` is the firebase command of course. I will change this in the README

theyokohamalife: @Anissa3005 What you'll need to do is update the command in package.json so it's `dev:startlocaldb --project="find-a-doc-japan"` and then remove the reference to it in the README.

Anissa3005: @theyokohamalife I have just added that, and it's working fine!

ermish (approved):
LGTM! :shipit:

PR #251, opened 2023-09-05
249/submission tests

### LINKED ISSUE #249: Add Tests for submissionService functions
- [x] Create **happy path** tests for the following:
   - [x] `getSubmissionById`
   - [x] `getSubmissions`
   - [x] `addSubmission`
   - [x] `updateSubmission`
- [x] Create **error path** tests for the following:
   - [x] `getSubmissionById`
   - [x] `addSubmission`
   - [x] `updateSubmission`
- [ ] Adding reset database for emulator

evan-desu:
Previously there were no unit tests for submissions. Now there are.

Resolves #249

# What changed

# Testing instructions


README.md
```diff
@@ -188,3 +188,24 @@ Then, just add breakpoints by clicking on the line number of your code.
 9. If you'd like to share the query you built, such as demonstrating how you tested your code, check out [Apollo Explorer's sharing features](https://www.apollographql.com/blog/announcement/platform/save-and-share-your-graphql-operations-in-apollo-explorer/#sharing-a-collection).
 
 </details>
+
+<details>
```
theyokohamalife, line 192: I'm a little confused about why these steps are needed. Generally, tests are something that we should be able to just run and not have to manually do anything in between.

evan-desu: @theyokohamalife This was something added to the other branch that came along when I created this new branch (still compared to `main`). Firebase requires these indexes for every possible combination of `fieldToOrder` and `orderDirection`. There is a way to automate their creation, but it looks a bit difficult to do. But I'm going to have a crack at it this morning and hopefully we won't need this section.

__tests__/submission.test.ts
```diff
@@ -0,0 +1,421 @@
...
+    }`,
+    variables: {
+        submissionId: '1'
+    }
+}
+
+describe('query getSubmissionById', () => {
+    let server: IMockServer
+
+    // TO DO: CLEAR (EMULATOR) DATABASE AFTER TESTS
```
Anissa3005, line 45: This is something we should add in a next PR, at the moment I get an error but could not figure out why. I need to look into it a bit more. 

Error:
**The host and port of the database emulator must be specified. (You may wrap the test script with `firebase emulators:exec './your-test-script'` to enable automatic discovery, or specify manually via initializeTestEnvironment({database: {host, port}}).**

__tests__/submission.test.ts
```diff
@@ -0,0 +1,421 @@
...
+    //         testEnvironment = await initializeTestEnvironment({
+    //         projectId: 'demo-test',
+    //         firestore: {host: 'localhost', port: 3004}
+    //     })
+    // })
+
+    // beforeEach(async () => {
+    //         await testEnvironment.clearDatabase()
+    // })
+
```
Anissa3005, line 57: This is something we should add in a next PR, at the moment I get an error but could not figure out why. I need to look into it a bit more.

Error:
**The host and port of the database emulator must be specified. (You may wrap the test script with firebase emulators:exec './your-test-script' to enable automatic discovery, or specify manually via initializeTestEnvironment({database: {host, port}}).**

__tests__/submission.test.ts
```diff
@@ -0,0 +1,421 @@
...
+import resolvers from '../src/resolvers'
+import { Submission } from '../src/typeDefs/dbSchema'
+import { ErrorCode } from '../src/result'
+import fs from 'fs'
+import path from 'path'
+// THIS IMPORT IS NEEDED WHEN YOU WANT TO CLEAR THE DATA FROM EMULATOR
+// import { initializeTestEnvironment, RulesTestEnvironment } from '@firebase/rules-unit-testing'
+// Please install this with @firebase/rules-unit-testing
+
+const schema = buildSchema(fs.readFileSync(
```
ermish, line 15: you can use `loadSchema()` from `schema.ts` file to load it instead of having to write it here

__tests__/submission.test.ts
```diff
@@ -0,0 +1,421 @@
...
+    path.join(__dirname, '../src/typeDefs/schema.graphql'),
+    'utf-8'
+))
```
ermish, line 18:
```suggestion
const schema = loadSchema()
```

ermish: you'll need to import the func

__tests__/submission.test.ts
Anissa3005: I have tried to implement this, but I encounter a type error.
The `addmockSchema` expects that `const schema` is type of `GraphQLSchema`. But the `loadSchema()` has type of DocumentNode. I'm still trying to find a solution but I doubt that this is possible (correct me if I'm wrong).

This is the error: 

_Type 'DocumentNode' is missing the following properties from type 'GraphQLSchema': description, extensions, astNode, extensionASTNodes, and 21 more._

_Type 'DocumentNode' is not assignable to type 'GraphQLSchema'._

theyokohamalife (approved):
LGTM 😎

PR #256, opened 2023-09-08
Add error messages/handeling to resolvers

### LINKED ISSUE #255: Add try and catch blocks for all resolvers
# Requirements

- [x] Add try catch block to all resolvers
- [x] Add custom errors in catch block

# Dependencies

- None

Anissa3005:
Previously, resolvers were missing try and catch blocks and this caused undesired behavior when making a query or mutation request when they missed required values/fields.

Resolves #255 

# What changed
1. Added try and catch blocks to all resolvers
2. Added the custom errors in the catch block

# Testing instructions

If you would make a mutation, example create submission, without an id as followed, it will throw an error now:

```mutation Mutation($input: SubmissionInput) {
    googleMapsUrl
    healthcareProfessionalName
    spokenLanguages {
      iso639_3
      nameEn
      nameJa
      nameNative
    }
  }
}
```

### DIFF
--- a/src/resolvers.ts
+++ b/src/resolvers.ts
@@ -94,7 +94,7 @@ const resolvers = {
             } catch (error) {
                 console.error(error)
                 console.log('INPUT_FIELD:', JSON.stringify(args))
-                return CustomErrors.missingInput('Failed to create facility with healthcare prof. Please make sure the input fields are correct')
+                return CustomErrors.missingInput('Failed to create facility and Healthcare Professional.')
             }  
         },
         createHealthcareProfessional: async (_parent: gqlType.HealthcareProfessional, args: {
@@ -116,7 +116,7 @@ const resolvers = {
             } catch (error) {
                 console.log(error)
                 console.log('INPUT_FIELD:', JSON.stringify(args))
-                return CustomErrors.missingInput('Failed to create the healthcare professional. Please make sure the input fields are correct.')
+                return CustomErrors.missingInput('Failed to create the healthcare professional.')
             }
         },
         createSubmission: async (_parent: gqlType.Submission, args: {
@@ -157,7 +157,7 @@ const resolvers = {
             } catch (error) {
                 console.log(error)
                 console.log('INPUT_FIELDS:', JSON.stringify(args))
-                return CustomErrors.missingInput('Failed to create submission. Please make sure the input fields are filled out correctly')
+                return CustomErrors.missingInput('Failed to create submission.')
             }
         },
         updateSubmission: async (_parent: gqlType.Submission, args: {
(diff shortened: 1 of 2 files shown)


theyokohamalife:
@Anissa3005 I recommend changing the title to something more descriptive. For example, "Improve error handling of ..."

src/resolvers.ts
```diff
@@ -2,41 +2,71 @@ import * as facilityService from './services/facilityService'
...
+                if (!args.id || !args.id.trim()) {
+                    throw new Error('An ID was not provided')
+                }
+                const matchingHealthcareProfessional = 
+                await healthcareProfessionalService.getHealthcareProfessionalById(args.id)
 
-            return matchingHealthcareProfessional
+                return matchingHealthcareProfessional        
+            } catch (error) {
+                return CustomErrors.notFound(`${error}`)
```
Anissa3005, line 45: This error message is less descriptive because the test is written in the way that it should be identical to the test.
I would like and change this with this issue: #262

src/resolvers.ts
```diff
@@ -2,41 +2,71 @@ import * as facilityService from './services/facilityService'
...
     Query: {
         facilities: async () => {
-            const matchingFacilities = await facilityService.searchFacilities(['1'])
+            try {
+                const matchingFacilities = await facilityService.searchFacilities(['1'])
 
-            return matchingFacilities
+                return matchingFacilities
+            } catch (error) {
+                return CustomErrors.notFound(`Failed to find facilities: ${error}`)
```
ermish, line 15: We don't actually want to return errors back to the user as it's a security risk and not helpful to them. 
We should 
1) log the error and user request for our own debugging (console.log is fine for now. I'll be moving all of them to real logs later), 
then 2) return a helpful message to the user. 

Let's update each of the try/catches with this. 
Consider what message you want to send and what action the user should take.

src/resolvers.ts
ermish: Always update your tests to match your code, not the other way around. Let's fix it in the issue you added ✅

ermish (changes requested):
Nice work! Added a comment

src/resolvers.ts
Anissa3005: By user request do you mean the arguments or the request body (contextValue) for example:
```
{
  "healthcareProfessionalId": "HhnkvZnpY3xWA7snihhr"
}
```

src/resolvers.ts
```diff
@@ -128,7 +190,9 @@ const resolvers = {
 
                 return updatedSubmission
             } catch (error) {
-                throw new Error(`Failed to update submission: ${error}`)
+                console.log(error)
+                console.log('INPUT_FIELDS:', JSON.stringify(args))
+                return CustomErrors.missingInput('Failed to update the submission.')
```
Anissa3005, line 195: Husky doesn't allow me to put a longer string for customerErrors so I had to keep it short

ermish: we can update eslint in a future ticket to allow longer strings. Personally I think strings can be really long if need be. It's not hard to scroll if we have to :)

ermish (approved):
Looks great @Anissa3005 ! 
Please update the tests to make sure they're passing our CI pipeline before we merge this

PR #263, opened 2023-09-12
Create `approveSubmission` function and logic

### LINKED ISSUE #220: Create an `updateSubmission` method
# Requirements

- [x] Create an `updateSubmission` method
- [ ] #296
    - [x] When a submission is approved, create a new facility and healthcareprofessionals from the submission data
        - [x] This should utilize the `add` methods in the corresponding `facilityService` and `healthcareProfessionalService` 
    - [x] When a submission is rejected, we don't delete it, just mark the flag `isRejected`. 
    - [ ] #297

### LINKED ISSUE #296: Approved Submissions should create a new Facility and/or HealthcareProfessional
# Summary

Let's create the ability to approve a submission. This will convert the submission into the appropriate facility and/or healthcare providers

# Requirements
When a Submission is approved, a couple different scenarios could happen:
1. If the Facility doesn't already exist, it should be created.
2. If the HealthcareProfessional doesn't exist, it should be created.
3. If the Facility exists, but the HealthcareProfessional doesn't, the HealthcareProfessional should be created and its ID added to the Facility.
4. If the HealthcareProfessional exists at another Facility, its ID should also be added to the Facility created by the Submission.

### LINKED ISSUE #318: Clean up `updateFacility` function
# Summary

Update `updateFacility` method to use the `Result` framework and clean up a few things. 

# Requirements

- [ ] Remove the extra `get` call at the end
- [ ] Return a `Result` object 
- [ ] Update the catch to return a `Result` object with an error code. (no stack trace included. log it instead)
- [ ] change return type to just be a boolean if it succeeded. No need to return an object

evan-desu:
# Summary

When a moderator flags a user submission as approved, a new facility and/or healthcare professionals will be created.

Resolves #220 
Resolves #318
Resolves #296 

**NOTE: tests were split into #323, facility<-->provider relationship updates were split into #324** 

# What changed
1. Added a conditional statement to `updateSubmission` to check is the submission has been marked as `approved`. If so, the `addFacility` and `addHealthcareProfessional` service functions are called.
2. Added a new table to schema called `UpdateSubmissionInput` in order to separate user inputs from moderator inputs.

# Testing instructions

- NOTE: this logic has not been tested yet to work. This will come later in #323. 
- For now, let's just validate the app still runs. Ensure `yarn dev` works and tests pass

### DIFF
--- a/src/services/submissionService.ts
+++ b/src/services/submissionService.ts
@@ -218,6 +218,7 @@ export const updateSubmission = async (submissionId: string, fieldsToUpdate: Par
         await submissionRef.set(updatedSubmissionValues, {merge: true})
 
         const updatedSubmission = await getSubmissionById(submissionRef.id)
+
         if (fieldsToUpdate.isApproved === true) {
             const facilityData: gqlTypes.FacilityInput = {
                 nameEn: submissionToUpdate.healthcareProfessionalName,
@@ -262,7 +263,7 @@ export const updateSubmission = async (submissionId: string, fieldsToUpdate: Par
             
             const newFacilityResult = await addFacility(facilityData)
             
-            if(newFacilityResult.data) {
+            if (newFacilityResult.data) {
                 healthcareProfessionalData.id = newFacilityResult.data.id
             } else {
                 throw new Error('Failed to create new facility.')
@@ -279,7 +280,7 @@ export const updateSubmission = async (submissionId: string, fieldsToUpdate: Par
     }
 }
 
-const validateSubmissionInputFields = (input: gqlTypes.SubmissionInput): Result<dbSchema.Submission> => {
+const validateSubmissionInputFields = (input: gqlTypes.AddSubmissionInput): Result<dbSchema.Submission> => {
     const validatedSubmissionResult: Result<dbSchema.Submission> = {
         hasErrors: false,
         errors: []
--- a/src/services/submissionService.ts
+++ b/src/services/submissionService.ts
@@ -6,14 +6,13 @@ import { CustomErrors, ErrorCode, Result } from '../result'
 import { hasSpecialCharacters } from '../../utils/stringUtils'
 import { addFacility } from './facilityService'
 import { addHealthcareProfessional } from './healthcareProfessionalService'
-import { satisfies } from 'semver'
 
 /**
  * Gets the Submission from the database that matches the id.
  * @param id A string that matches the id of the Firestore Document for the Submission.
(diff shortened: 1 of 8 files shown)


src/services/submissionService.ts
```diff
@@ -175,6 +175,53 @@ export const updateSubmission = async (id: string, updatedFields: Partial<dbSche
...
+                        addressLine2En: '',
+                        prefectureJa: '',
+                        cityJa: '',
+                        addressLine1Ja: '',
+                        addressLine2Ja: ''
+                    }
+                },
+                healthcareProfessionals: [],
+                isDeleted: false
+            }
```
evan-desu, line 201: @ermish @theyokohamalife 
This is my main area of concern. Currently, when we pass the submission ID and mark the submission as `isApproved`, a new facility and healthcarePro are created, which is great. However, there are several fields which are hardcoded, and only the fields from `AddSubmissionInput` are added to the the new facility/healthcarePro (`googleMapsUrl`, `healthcareProfessionalName`, and `spokenLanguages`).

In my mind, the user completes the basic `addSubmission` form with the three `AddSubmissionInput` fields, the moderator/admin approves it, and the submission gets added to the DB as a new facility/healthcarePro. But what about the other fields (contact, specialty, etc.)? Do they come form the admin who approves? Or is the user also meant to include this extra info? Any guidance or advice would be mush appreciated 🙏

theyokohamalife: @evan-desu The other fields such as contact, specialty, etc are entered by the moderator after confirming the information with the healthcare provider.

evan-desu: @theyokohamalife Thank you for confirming that for me!

evan-desu: After rebasing, `updateSubmission` no longer adds a facility or healthcarePro, most likely due to changes made to `addFacility` and/or `addHealthcareProfessional`.

ermish: update: I've changed the logic to be where the moderator should fill out the facility/provider before approving. This will prevent incomplete results to show up in user searches

ermish:
hmmm....for whatever reason, the changes in this PR don't seem to match. It's showing changes that were made in main branch. Going to close this and reopen a new PR and see if it changes.

ermish:
looks like the issue persists in a new one....re-opening

theyokohamalife (approved):
Approving this after our mutual decision to fix CI together with the remaining service functions in the next PR.

PR #264, opened 2023-09-12
Refactor the error message tests

### LINKED ISSUE #262: Change the healthcare professional error path test so the error message is more dynamic
# Requirements

- [ ] Fix error message test

# Dependencies

- None

Anissa3005:
Previously, the test only passed if the error message matched the tested sentence exactly. To make a more dynamic I refactor the way we test the error message.

Resolves #262 

# What changed
Refactor the tests for the error path, focusing more on the error message. Add a spyOn mock function to verify if Customerrors.notFound() has been called. Also, test the error message by checking if it is defined and has a length greater than 0, rather than a specific message.

### DIFF
--- a/__tests__/server.test.ts
+++ b/__tests__/server.test.ts
@@ -106,8 +106,6 @@ describe('query healthcareProfessionalById', () => {
         const mySchema = addResolversToSchema({schema, resolvers})
         const preserveResolvers = true
         const mocks = {}
-
-        const spyCustomeErrors = jest.spyOn(CustomErrors, 'notFound')
         
         server = mockServer(mySchema, mocks, preserveResolvers)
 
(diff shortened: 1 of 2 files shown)


__tests__/server.test.ts
```diff
@@ -111,6 +118,8 @@ describe('query healthcareProfessionalById', () => {
         expect(errors[0].extensions.http).toEqual({status: 404})
         expect(errors.length).toBe(1)
         expect(errors[0].extensions.code).toBe('NOT_FOUND')
-        expect(errors[0].message).toBe('Healthcare professional not found.')
+        expect(errors[0].message).toBeDefined()
+        expect(errors[0].message.length).toBeGreaterThan(0)
+        expect(spyCustomeErrors).toHaveBeenCalled()
```
ermish, line 123: Are you planning to check the message back to the user in this PR or in a future one?

Anissa3005: So, this fix was because of issue #255 not passing. I initially believed that my error messages should be more dynamic. However, after reading your feedback (in PR that resolves #255), I'm starting to think that this might not have been necessary at all.

My question now is, what is the best way to test for error messages: as it was done before, or a combination of both?

ermish: hmmm...I think that since these are endpoint/end-to-end/api/whateveryoucallthem tests, we should just test what is returned back to the user and not take the spy approach

ermish (approved):
ship it! :shipit:

PR #265, opened 2023-09-15
Require all fields for addSubmission function

### LINKED ISSUE #261: fix: createSubmissions Service Functions
# Summary
Currently, a user is able to create a submission without including a `googleMapsUrl`, `healthcareProfessionalName`, or `spokenLanguages`. We want to prevent users from doing this.

# Requirements
- [ ] Add logic to require users to include all fields.

evan-desu:
Previously, submissions with empty fields were accepted and added to the database. This ensures that all required fields (googleMapsUrl, healthcareProfessionalName, and spokenLanguages) are provided.

Resolves #261 

# What changed
1. Added validation to ensure` googleMapsUrl` and `healthcareProfessionalName` are not empty.
2. Validated `spokenLanguages` and ensured it's not an empty array.

# Testing instructions
Copy and paste this into the GraphQL Studio:

**Operation:**
```
mutation Mutation($input: SubmissionInput) {
    id
      }
}
```

**Variables:**
```
{
  "input": {
    "googleMapsUrl": " ",
    "healthcareProfessionalName": "Dr Bob",
    "spokenLanguages": [
      {
        "iso639_3": "eng",
        "nameJa": "English",
        "nameEn": "English",
        "nameNative": "English"
      }
    ]
  }
}
```
You should receive an error similar to this:
```
{
  "data": {
    "createSubmission": null
  },
  "errors": [
    {
      "message": "Error adding submission: Error: Fields cannot be empty strings.",
      "locations": [
        {
          "line": 3,
          "column": 3
        }
      ],
```

### DIFF
--- a/src/resolvers.ts
+++ b/src/resolvers.ts
@@ -2,7 +2,7 @@ import * as facilityService from './services/facilityService'
 import * as healthcareProfessionalService from './services/healthcareProfessionalService'
 import * as gqlType from './typeDefs/gqlTypes'
 import * as submissionService from './services/submissionService'
-import CustomErrors from './errors'
+import { CustomErrors } from './result'
 
 const resolvers = {
     Query: {
@@ -133,7 +133,7 @@ const resolvers = {
         }) => {
             try {
                 for (const value of Object.values(args.input)) {
-                    if (!value) {
+                    if (value === null || value === undefined) {
                         throw new Error('Missing Input')
                     }
                 }
@@ -156,9 +156,12 @@ const resolvers = {
                     createdDate: new Date().toISOString(),
                     updatedDate: new Date().toISOString()
                 }
-                const newSubmission = await submissionService.addSubmission(submissionData)
-    
-                return newSubmission
+                const addSubmissionResult = await submissionService.addSubmission(submissionData)
+
+                if (addSubmissionResult.hasErrors) {
+                    throw new Error('Failed to create submission.')
+                }
+                return addSubmissionResult.data
             } catch (error) {
                 console.log(error)
                 console.log('INPUT_FIELDS:', JSON.stringify(args))
@@ -177,9 +180,17 @@ const resolvers = {
             }
         }) => {
             try {
-                const updatedSpokenLanguages = args.input.spokenLanguages ?
-                    submissionService.mapAndValidateSpokenLanguages(args.input.spokenLanguages)
-                    : undefined
+                let updatedSpokenLanguages: gqlType.SpokenLanguage[] | undefined
+                
+                if (args.input.spokenLanguages) {
+                    const spokenLanguagesResult = submissionService
+                        .mapAndValidateSpokenLanguages(args.input.spokenLanguages)
+                    
+                    if (spokenLanguagesResult.hasErrors) {
+                        throw new Error('Invalid spoken Language provided.')
+                    }
+                    updatedSpokenLanguages = spokenLanguagesResult.data
+                }
                 
                 const gqlUpdatedFields: Partial<gqlType.Submission> = {
                     ...args.input,
--- a/src/services/submissionService.ts
+++ b/src/services/submissionService.ts
@@ -2,6 +2,7 @@ import { DocumentData, Query } from 'firebase-admin/firestore'
 import * as gqlType from '../typeDefs/gqlTypes'
 import * as dbSchema from '../typeDefs/dbSchema'
 import { dbInstance } from '../firebaseDb'
+import { ErrorCode, Result } from '../result'
 
 export const getSubmissionById = async (id: string) : Promise<dbSchema.Submission | null> => {
     try {
@@ -116,23 +117,47 @@ function convertToDbSubmission(submission: gqlType.SubmissionInput):
 }
 
 export const addSubmission = async (submissionInput: gqlType.SubmissionInput): 
-    Promise<dbSchema.Submission> => {
-    try {
-        const dbSubmission = convertToDbSubmission(submissionInput)
-
-        const submissionRef = dbInstance.collection('submissions')
-
-        const docRef = await submissionRef.add(dbSubmission)
-
-        const newSubmission: dbSchema.Submission = {
-            ...dbSubmission,
-            id: docRef.id
-        }
-
-        return newSubmission
-    } catch (error) {
-        throw new Error(`Error adding submission: ${error}`)
+    Promise<Result<dbSchema.Submission>> => {
+    const addSubmissionResult: Result<dbSchema.Submission> = {
+        hasErrors: false,
+        errors: []
     }
+
+    const validationResults = validateSubmissionInputFields(submissionInput)
+
+    if (validationResults.hasErrors) {
+        return validationResults
+    }
+
+    const nonNullLanguages = submissionInput.spokenLanguages
+        .filter(lang => !!lang) as gqlType.SpokenLanguageInput[]
+
+    const spokenLanguagesResult = mapAndValidateSpokenLanguages(nonNullLanguages)
+
+    if (spokenLanguagesResult.hasErrors) {
+        return {
+            hasErrors: true,
+            errors: spokenLanguagesResult.errors
+        }
+    }
+
+    const dbSubmission = convertToDbSubmission({
+        ...submissionInput,
+        spokenLanguages: spokenLanguagesResult.data as gqlType.SpokenLanguageInput[]
+    })
+
+    const submissionRef = dbInstance.collection('submissions')
+
+    const docRef = await submissionRef.add(dbSubmission)
+
+    const newSubmission: dbSchema.Submission = {
+        ...dbSubmission,
+        id: docRef.id
+    }
+
+    addSubmissionResult.data = newSubmission
+
+    return addSubmissionResult
 }
 
 export const updateSubmission = async (id: string, updatedFields: Partial<dbSchema.Submission>): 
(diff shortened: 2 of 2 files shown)


src/services/submissionService.ts
```diff
@@ -118,7 +118,23 @@ function convertToDbSubmission(submission: gqlType.SubmissionInput):
 export const addSubmission = async (submissionInput: gqlType.SubmissionInput): 
     Promise<dbSchema.Submission> => {
     try {
-        const dbSubmission = convertToDbSubmission(submissionInput)
+        if (!submissionInput.googleMapsUrl.trim() || !submissionInput.healthcareProfessionalName.trim()) {
```
ermish, line 121: Let's move all the validation work into a dedicated validation function

ermish: Should check for null/undefined before calling trim

src/services/submissionService.ts
```diff
@@ -118,7 +118,23 @@ function convertToDbSubmission(submission: gqlType.SubmissionInput):
...
+            throw new Error('Submission fields cannot be empty strings.')
+        }
+
+        const nonNullLanguages = submissionInput.spokenLanguages
+            .filter(lang => lang !== null) as gqlType.SpokenLanguageInput[]
```
ermish, line 126: Null is actually different then undefined. We should check for both. 
You can likely use `!!lang` for the check

src/services/submissionService.ts
```diff
@@ -118,7 +118,23 @@ function convertToDbSubmission(submission: gqlType.SubmissionInput):
...
+
+        const validatedLanguages = mapAndValidateSpokenLanguages(nonNullLanguages)
+
+        if (!validatedLanguages || validatedLanguages.length === 0) {
```
ermish, line 130: Move this into the map and validate function. 
This is probably a good time for us to introduce a proper errors object. 
Have it return a `hasErrors` Boolean and and errors array where each error is an object with a `message` field and `property` field
Example

```
result: {
    hasErrors: true,
    errors: [
        { field: "spokenlanguages[1].type",
          message: "the type can only be english or Japanese" }]
}
```

ermish (changes requested):
We can talk on slack if you need help with some of this. 
It's a new design

src/services/submissionService.ts
ermish: update: 
1) the error message should be based on the custom errors type.
2) we should make this result object a typscript type that can be used across all the services.

ermish: update #2: I have a PR with a results framework set up that should make this a bit easier and more consistent.

theyokohamalife (approved):
LGTM!

.env.prod
```diff
@@ -7,6 +7,4 @@ FIRESTORE_PROJECT_ID = "find-a-doc-japan"
 # https://console.firebase.google.com/u/0/project/find-a-doc-japan/overview
 # DO NOT COMMIT THESE VALUES TO GIT!!!
 
-# FIRESTORE_SERVICE_ACCOUNT = 'enteryourkeyhere'
-
-
+# FIRESTORE_SERVICE_ACCOUNT = 'enteryourkeyhere'
```
theyokohamalife, line 10: I think you can leave this out since it's already on main.

evan-desu: @theyokohamalife I'll be sure to remove it (or figure out how to do that).

src/services/submissionService.ts
```diff
@@ -158,6 +168,16 @@ export const updateSubmission = async (id: string, updatedFields: Partial<dbSche
     }
 }
 
+const validateSubmissionInputFields = (input: gqlType.SubmissionInput) => {
+    if (
+        (input.googleMapsUrl === null || input.googleMapsUrl === undefined || !input.googleMapsUrl.trim()) ||
```
ermish, line 173:
```suggestion
        (input.googleMapsUrl || !input.googleMapsUrl.trim()) ||
```

didn't see this sooner. you can simplify a null and undefined check with a basic truthy check. (truthy checks for both and empty strings, but you need to be careful if things like 0 are valid options.

src/services/submissionService.ts
```diff
@@ -158,6 +168,16 @@ export const updateSubmission = async (id: string, updatedFields: Partial<dbSche
...
+        (input.healthcareProfessionalName === null || input.healthcareProfessionalName === undefined ||
```
ermish, line 174:
```suggestion
        (input.healthcareProfessionalName || 
```

src/services/submissionService.ts
```diff
@@ -196,7 +216,9 @@ function gqlSpokenLanguageToDbSpokenLanguage(lang: gqlType.SpokenLanguage):
 
 export const mapAndValidateSpokenLanguages = (spokenLanguages: gqlType.SpokenLanguageInput[]): 
     dbSchema.SpokenLanguage[] | undefined => {
-    if (!spokenLanguages || ! spokenLanguages.length) { return undefined }
+    if (!spokenLanguages || spokenLanguages.length === 0) {
```
ermish, line 219:
```suggestion
    if (!spokenLanguages) {
```

src/resolvers.ts
```diff
@@ -133,7 +133,7 @@ const resolvers = {
         }) => {
             try {
                 for (const value of Object.values(args.input)) {
-                    if (!value) {
```
ermish, line 136: Wait. This should be !value. It's the equivalent.

src/resolvers.ts
```diff
@@ -157,8 +157,11 @@ const resolvers = {
                     updatedDate: new Date().toISOString()
                 }
                 const newSubmission = await submissionService.addSubmission(submissionData)
-    
-                return newSubmission
+
+                if (newSubmission.hasErrors) {
+                    throw new Error('Failed to create submission.')
+                }
+                return newSubmission.data
```
ermish, line 164: Perfect!

src/resolvers.ts
```diff
@@ -157,8 +157,11 @@ const resolvers = {
                     updatedDate: new Date().toISOString()
                 }
                 const newSubmission = await submissionService.addSubmission(submissionData)
```
ermish, line 159: nit: addSubmissionResult would be a more accurate name.

src/resolvers.ts
```diff
@@ -177,9 +180,17 @@ const resolvers = {
...
+                
+                if (args.input.spokenLanguages) {
+                    const spokenLanguagesResult = submissionService
+                        .mapAndValidateSpokenLanguages(args.input.spokenLanguages)
+                    
+                    if (spokenLanguagesResult.hasErrors) {
+                        throw new Error('Invalid spoken Language provided.')
+                    }
+                    updatedSpokenLanguages = spokenLanguagesResult.data
+                }
```
ermish, line 193: This is ok for now. 
@evan-desu  can you create a new issue to move all of the mapping and validation work into the service? 
The idea is to keep your resolvers as simple as possible and have the service method do all the work.

src/services/submissionService.ts
```diff
@@ -158,6 +183,43 @@ export const updateSubmission = async (id: string, updatedFields: Partial<dbSche
     }
 }
 
+const validateSubmissionInputFields = (input: gqlType.SubmissionInput): Result<dbSchema.Submission> => {
+    const validatedSubmissionResult: Result<dbSchema.Submission> = {
+        hasErrors: false,
+        errors: []
+    }
+    
+    if (input.googleMapsUrl === null || input.googleMapsUrl === undefined || !input.googleMapsUrl.trim()) {
```
ermish, line 192: Use shorthand 😎

```suggestion
    if (!input.googleMapsUrl || !input.googleMapsUrl.trim()) {
```

src/services/submissionService.ts
```diff
@@ -158,6 +183,43 @@ export const updateSubmission = async (id: string, updatedFields: Partial<dbSche
...
+        validatedSubmissionResult.hasErrors = true
+        validatedSubmissionResult.errors?.push({
+            field: 'googleMapsUrl',
+            errorCode: ErrorCode.MISSING_INPUT,
+            httpStatus: 400
+        })    
+    }
+
+    if (input.healthcareProfessionalName === null || input.healthcareProfessionalName === undefined ||
```
ermish, line 201:
```suggestion
    if (!input.healthcareProfessionalName || 
```

src/services/submissionService.ts
```diff
@@ -158,6 +183,43 @@ export const updateSubmission = async (id: string, updatedFields: Partial<dbSche
...
+        validatedSubmissionResult.hasErrors = true
+        validatedSubmissionResult.errors?.push({
+            field: 'googleMapsUrl',
+            errorCode: ErrorCode.MISSING_INPUT,
+            httpStatus: 400
+        })    
+    }
+
+    if (input.healthcareProfessionalName === null || input.healthcareProfessionalName === undefined || 
+        !input.healthcareProfessionalName.trim()) {
```
ermish, line 202: We should put a limit on all these string lengths. 256 should be more than enough

ermish: N/m ha just saw the next one. 

We should probably limit the map url to 1028

ermish (approved):
Great! Logic looks sound. Just some small optimizations and we're good to go

.env.prod
theyokohamalife: @evan-desu  You should be able to run `git checkout main .env.prod` and then commit and push to have this removed.

evan-desu: @theyokohamalife I'll try to do that later this evening. I gave up yesterday 😅

ermish (approved):
Ship it! :shipit:

PR #269, opened 2023-09-19
reworked facilities service

### LINKED ISSUE #222: Create the `getFacilities` method
# Requirements

- [ ] Create a `getFacilities` method
    - [ ] Add parameters to be able to query by commonly used fields such as `location` in the facility or `spokenLanguages`, `specialties`, under the list of healthcare professionals.  (query tip: might want to consider doing the query for healthcare professionals first, then pull the associated facilities)  
    - [ ] (optional) Add logic to orber by
        - [ ] User can send in an array of "orderby" objects. The orderby object contains: `orderDirection` (options can be: "ascending" or "descending"), `fieldToOrder`.
        - [ ] Add the logic in the firebase query to order by. (You'lll likely need to translate our object to firebase style object in the query)
    - [ ] Add the limit logic
        - [ ] Add a paramter to limit the number of results returned. 
            - [ ]  Name this `limit`

[Helpful firebase link for order by and limit](https://firebase.google.com/docs/firestore/query-data/order-limit-data)

### LINKED ISSUE #224: Fix `getFacilityById`

### LINKED ISSUE #266: Build new API response framework
# Summary

We want our API to have a consistent format in how it returns data when successful and helpful errors when something goes wrong. 

# Requirements

- [ ] Create a standardized response object for the API
- [ ] Create a result object for use throughout the services
- [ ] Build an error message framework
    - [ ] support consistent messages
    - [ ] validation
    - [ ] error checking
    - [ ] i18n support. (use error codes instead of strings)
- [ ] Lay the groundwork for i18n support for messages from the server. The server side will send codes that the client-side maps to translation files

### LINKED ISSUE #267: Validate input fields for facilities functions
# Summary

We need to validate the parameters that users are sending in. This should be for all queries and mutations. 
This ticket will focus on the facilities based methods. 

# Requirements

- [x] Ensure the values sent in are valid.
    - [x] type checking should already be covered by graphql
- [ ] ensure dynamic fields (like sort by - field name) matches valid options. Ex. can you sort by id? (we shouldn't) what about "abc"? (field doesn't exist)
- [ ] Put limits on number fields. limit for example probably shouldn't go over 1000. if a user sends in -1 or 999999999, it should be invalid. 
- [ ] Check for empty strings

### Specific to queries
- [ ] Check for malicious strings like js being injected as a string to be ran. (we can use a library, no pref)
- [ ]  sort-by should be limited to two columns max.  (future ticket: we add this to our api documentation)

### Specific to Mutations

- [ ] Which fields should not be valid to add/update?
    - [ ] id, created date, and updated date are some examples that shouldn't be editable.

### LINKED ISSUE #270: Implement pagination for querying facilities.
# Summary

Implement pagination for querying facilities.

# Requirements

- [ ] add the ability to limit the number of results returned
- [ ] add the ability to skip or offset the number of results. This is the equivalent of pages. 
- [ ] add default values of limit = 20 and skip = 0

ermish:
Resolves #222 
Resolves #224 
Resolves #266 
Resolves #267 
Resolves #270

# What changed

- [x] Created a new Results framework. This enables us to better handle errors in the server-side, how to know when to stop if validation fails, giving the user a better error message, and providing consistency for our gql API responses. 🔥 
- [x] Reworked Facilities service 🚧 🏗️ 
    - [x] Rebuilt `getfacilitiesbyid`
        - [x] Added validation ✨ 
        - [x] Added in result design
        - [x] fixed query issue
    - [x] Rebuilt `searchFacilities`/`getFacilities`
        - [x] Added a bunch of validation logic ✨ 
        - [x] Built out new query system. Similar design to submission sec (shoutout @evan-desu for saving me time 🙌 )
        - [x] Added in default values for limit/orderby
        - [x] Added limits to strings and such, validated emails.
        - [x] solved a ton of type issues
    - [x] Rebuilt the `addFacility` function 👷   
        - [x] Reworked all the logic
        - [x] 🦖  Added in new types and migrated to work more with gql types instead of dbTypes. The advantage is now we don't need to clone the object and we can use the user input directly after it's been validated.  
        - [x] Had to rework the `addHealthcareProvider` function as a side-effect/dependency for facility. 🚧 
            - [x] ✅ Improved the type issues
            - [x] ✅ Added in the new result framework
            - [ ] ❌ Skipped over adding validation for a separate PR
            - [x] ✅  You can now save multiple healthcare providers with one facility! 🥳  
- [x] Add pagination capabilities for querying facilities 📖 
- [x] Solved a bunch of various type issues (ids, missing fields, null fields, etc) 😌 
- [x] Created string utils 🧶 
- [x] Updated mock data to work with updated types and new facilities service functions 🏘️ 

# Testing instructions

1. Test out the `getFacilityById` function using the facility query. 
2. Send in proper id.

```
query Facility($facilityId: ID!) {
  facility(id: $facilityId) {
    contact {
      address {
        addressLine1En
        addressLine1Ja
        addressLine2En
      }
      id
      email
      mapsLink
      phone
      website
    }
  }
}

Variables
{
  "facilityId": "youridhere"
}
```

3. Test out the `searchFacilities` function using the facilities query. 
4. Test all the various filter options (see gql schema FacilitiesSearchFilters for options)

Use this query as a starting point

```
query QueryFacilities($filters: FacilitySearchFilters) {
  facilities(filters: $filters) {
    id
    contact {
      address {
        addressLine1En
        addressLine2En
        addressLine1Ja
        cityJa
        cityEn
      }
      email
      id
      mapsLink
      phone
      website
    }
    isDeleted
    nameEn
    nameJa
    healthcareProfessionals {
      acceptedInsurance
      degrees {
        nameJa
        nameEn
      }
      names {
        firstName
        lastName
        locale
        middleName
      }
      specialties {
        names {
          locale
          name
        }
      }
      spokenLanguages {
        nameJa
        nameNative
        nameEn
        iso639_3
      }
    }
    healthcareProfessionalIds
  }
}


Variables
{
  "filters": {
    "limit": 10,
    "offset": 0,
    "nameEn": "Zoo"
  }
}

```


src/firebaseDb.ts
```diff
@@ -64,7 +64,7 @@ export const initiatilizeFirebaseInstance = async () => {
     } else if (isTestingEnvironment || isLocal) {
         console.log('\n🔥 Connecting to firebase emulator...')
         console.log('TIP: if it doesn\'t connect after 10 secs,' +
-            'make sure you have the firebase emulator running using the "yarn dev" command')
+            'make sure you have the firebase emulator running using the "yarn dev:startlocaldb" command')
```
theyokohamalife, line 67:
```suggestion
            ' make sure you have the firebase emulator running using the "yarn dev:startlocaldb" command')
```

theyokohamalife: Nit: Need a space after the comma

Screenshot 2023-09-20 at 3 53 27 PM

theyokohamalife (approved):
Made one small suggestion, but otherwise LGTM

PR #306, opened 2023-10-03
Test/create healthcare professional

theyokohamalife:
Resolves #[Issue number]

# What changed

# Testing instructions


__tests__/healthcareProfessional.test.ts
```diff
@@ -0,0 +1,199 @@
...
+import resolvers from '../src/resolvers'
+import loadSchema from '../src/schema'
+import request from 'supertest'
+import { ApolloServer } from '@apollo/server'
+import { startStandaloneServer } from '@apollo/server/standalone'
+import { initiatilizeFirebaseInstance } from '../src/firebaseDb'
+
+const facilityIds = [] as string[]
+
+const facilityQueryData = {
```
theyokohamalife, line 11: Ideally, we'd like a factory to create a Facility for us so this wouldn't be needed.

__tests__/healthcareProfessional.test.ts
```diff
@@ -0,0 +1,211 @@
...
+            ],
+            isDeleted: null,
+            nameEn: null,
+            nameJa: null
+        }
+    }
+
+}
+
+const healthcareProfessionalQueryData = {
```
ermish, line 85: nit: we should be more specific with what type of query this is. Passing, failing, with languages, etc.

__tests__/healthcareProfessional.test.ts
```diff
@@ -0,0 +1,211 @@
...
+describe('createHealthcareProfessional', () => {
+    let url: string
+
+    const server = new ApolloServer({
+        typeDefs: loadSchema(),
+        resolvers
+    })
+    
+    beforeAll(async () => {
+        ({ url } = await startStandaloneServer(server, {listen: { port: 0 }}))
```
ermish, line 168: is port 0 ok here?

__tests__/healthcareProfessional.test.ts
```diff
@@ -0,0 +1,211 @@
...
+        // Create a new Facility to add HealthProfessionals to
+        const facility = await request(url).post('/').send(facilityQueryData)
+
+        const facilityId = await facility.body.data.createFacilityWithHealthcareProfessional.id
+
+        facilityIds.push(facilityId)
+    })
+    
+    afterAll(async () => {
+        await server?.stop()
```
ermish, line 180: if we don't clear the test data we added, we can end up with a bunch of the same data over time. This should be fixed in the future.

src/services/healthcareProfessionalService.ts
```diff
@@ -55,51 +55,67 @@ export async function addHealthcareProfessional(
...
+ */
+export async function addHealthcareProfessionalToFacility( 
+    healthcareProfessionalInput: gqlTypes.HealthcareProfessionalInput
+) {
+    const addHealthcareProfessionalResult : Result<dbSchema.HealthcareProfessional> = {
+        hasErrors: false,
+        errors: []
+    }
+
+    if (healthcareProfessionalInput.facilityIds.length == 0) {
```
ermish, line 71:
```suggestion
    if (!healthcareProfessionalInput.facilityIds.length) {
```

src/services/healthcareProfessionalService.ts
```diff
@@ -55,51 +55,67 @@ export async function addHealthcareProfessional(
...
+        throw CustomErrors.missingInput('The list of facilityIds cannot be empty.')
```
ermish, line 72: we should migrate off the custom errors to using the result object with the error code. This will let us have localized language error messages. 
could add an error code like `ADDHEALTHCAREPROF_FACILITYIDS_REQUIRED`

src/services/healthcareProfessionalService.ts
```diff
@@ -55,51 +55,67 @@ export async function addHealthcareProfessional(
...
 
-// export async function searchHealthcareProfessionals(userSearchQuery : string[]) {
-// TODO: make it filter by params
-// const db = getFirestore()
-// const healthcareProfessionalRef = db.collection('healthcareProfessionals')
-// const snapshot = await healthcareProfessionalRef.where('id', 'in', userSearchQuery).get()
+        const facilities = healthcareProfessionalInput.facilityIds
 
-// const healthcareProfessionals = [] as HealthcareProfessional[]
+        facilities.map(async facilityId => {
```
ermish, line 86:
```suggestion
        for await (const facilityId of facilities) {
```

this `map(async` doesn't guarantee all the threads will return and could cause collisions. `for await` is the easiest way to iterate on async operations.

src/services/healthcareProfessionalService.ts
```diff
@@ -55,51 +55,67 @@ export async function addHealthcareProfessional(
...
+            const facilityRef = dbInstance.collection('facilities').doc(facilityId)
+                
+            facilityRef.update(
```
ermish, line 89:
```suggestion
            await facilityRef.update(
```

src/services/healthcareProfessionalService.ts
```diff
@@ -55,51 +55,67 @@ export async function addHealthcareProfessional(
...
 }
 
-export async function addHealthcareProfessionalToFacility(facilityId: string, 
-    healthcareProfessional: gqlTypes.HealthcareProfessionalInput) {
+/**
+ * Creates a HealthcareProfessional and adds it to the listed facilities
+ * @param healthcareProfessionalInput 
+ * @returns A HealthcareProfessional object
+ */
+export async function addHealthcareProfessionalToFacility(
```
ermish, line 63: I think this function might actually be no longer necessary. Let's talk through it on slack

__tests__/healthcareProfessional.test.ts
theyokohamalife: Yup, port `0` was intentional so we don't have to worry about it interfering with dev or prod environments.

__tests__/healthcareProfessional.test.ts
theyokohamalife: Since we don't have any examples of that in the repo, please use the `suggestion` button to suggest the specific code change you want and I'll make sure the other tests get updated as well. Otherwise, we could end up going back and forth because I'm not exactly sure what you're asking for.

ermish (approved):
ship it 🚀

PR #310, opened 2023-10-07
update max length to ignore strings and comments

ermish:
# What changed

Quick commit to ignore the max length rule for strings, urls, comments, and non-code related lines. 
We live in a world above 720p resolution monitors ✨


theyokohamalife (approved):
LGTM

PR #325, opened 2023-10-15
Rework schema, inputs, and tests

### LINKED ISSUE #278: Move Mapping and Validation Work into the updateSubmission in Service
# Summary
In order to to keep the resolvers as simple as possible, we need to move all the mapping and validation work for `updateSubmission` from the resolver to the service method.

# Requirements

- [ ] Move the logic

### LINKED ISSUE #303: Implement updateHealthcareProfessional
# Summary
We need a way to update a HealthcareProfessional in the database.

# Requirements

- [ ] Write the `updateHealthcareProfessional` resolver
- [ ] Write the `updateHealthcareProfessional` mutation
- [ ] Add happy path test for `updateHealthcareProfessional`

### LINKED ISSUE #314: BUG: firestore isn't clearing before the test is done
# Summary

there's an operation that doesn't finish before jest shuts down. If you add the flag `--detectOpenHandles` to the `yarn test` command, an error will pop up

``` terminal
Test Suites: 8 passed, 8 total
Tests:       32 passed, 32 total
Snapshots:   0 total
Time:        12.402 s
Ran all test suites.

Jest has detected the following 1 open handle potentially keeping Jest from exiting:

  ●  TCPWRAP

      307 |
      308 |     beforeEach(async () => {
    > 309 |         await testEnv.clearFirestore()
          |                       ^
      310 |     })
      311 |
      312 |     afterAll(async () => {

      at node_modules/@firebase/rules-unit-testing/node_modules/node-fetch/lib/index.js:1468:15
      at Object.fetch [as default] (node_modules/@firebase/rules-unit-testing/node_modules/node-fetch/lib/index.js:1437:9)
      at RulesTestEnvironmentImpl.<anonymous> (node_modules/@firebase/rules-unit-testing/src/impl/test_environment.ts:109:29)
      at step (node_modules/@firebase/rules-unit-testing/dist/index.cjs.js:74:23)
      at Object.next (node_modules/@firebase/rules-unit-testing/dist/index.cjs.js:55:53)
      at node_modules/@firebase/rules-unit-testing/dist/index.cjs.js:48:71
      at __awaiter (node_modules/@firebase/rules-unit-testing/dist/index.cjs.js:44:12)
      at RulesTestEnvironmentImpl.Object.<anonymous>.RulesTestEnvironmentImpl.clearFirestore (node_modules/@firebase/rules-unit-testing/dist/index.cjs.js:337:16)
      at Object.<anonymous> (__tests__/submissions.test.ts:309:23)
```

# Requirements

- [ ] Add `--detectOpenHandles` to `yarn test`
- [ ] Fix the bug

### LINKED ISSUE #319: Create distinct add and update inputs for facility functions
# Summary

Currently, our api uses shared types that don't have the appropriate fields that are allowed to be added/updated. We should make the inputs specific to what is allowed for the action. 

For example, an `update` action shouldn't have be able to change createdDate or id.

# Requirements

- [ ] Create unique add input gqlType for `addFacility`
- [ ] Create unique update input gqlType for `updateFacility`
- [ ] Delete should just be an `id`

### LINKED ISSUE #320: Create distinct add and update inputs for healthcareprofessional functions
# Summary

Currently, our api uses shared types that don't have the appropriate fields that are allowed to be added/updated. We should make the inputs specific to what is allowed for the action. 

For example, an `update` action shouldn't have be able to change createdDate or id.

# Requirements

- [ ] Create unique add input gqlType for `addHealthcareProfessional`
- [ ] Create unique update input gqlType for `updateHealthcareProfessional`
- [ ] Delete should just be an `id`

### LINKED ISSUE #321: Create distinct add and update inputs for submission functions
# Summary

Currently, our api uses shared types that don't have the appropriate fields that are allowed to be added/updated. We should make the inputs specific to what is allowed for the action. 

For example, an `update` action shouldn't have be able to change createdDate or id.

# Requirements

- [ ] Create unique add input gqlType for `addSubmission`
- [ ] Create unique update input gqlType for `updateSubmission`
- [ ] Delete should just be an `id`

### LINKED ISSUE #322: Update tests to use new add/update Input gqlTypes
# Summary

We're making a lot of changes for the service inputs and outputs. Let's update the tests to match. 

# Requirements

- [ ] Update all facility/healthcareprovider/submission tests to use proper add/update input gqlTypes 
- [ ] Ensure they are using the Result object properly

# Dependencies

- #319 
- #320 
- #321 
- #296

### LINKED ISSUE #323: Create approveSubmission test
# Summary

Approve submission has a lot of logic. let's make some tests to cover key scenarios

# Requirements

- [ ] Test scenario 1: Add happy path test to approve submission. 
    - [ ] Should create a facility 
    - [ ] Should create healthcare providers 
    - [ ] Should create the connection between facility<->healthcare providers
- [ ] Test scenario 2: We cannot approve/reject an already approved submission
    - [ ] approve, then try to approve again
    - [ ] reject, then try to approve
    - [ ] approve, then try to reject

# Dependencies

- #296

### LINKED ISSUE #345: Unify gql schema and db schema
# Summary

Our schemas are effectively the same between the db and gql api. While they may diverge over time, we should try to share the same models/fields when possible

# Requirements

- [ ] update the db schema to be based on the gql api as a default. 
- [ ] Use `Omit<>` type to remove any sensitive fields

# Dependencies

### LINKED ISSUE #346: Rework `createHealthcareProfessional`
# Summary

The create method could use some love
- Add/improve validation
- Update gql->db mapping
- simplify
- Add error and result frameworks

# Requirements

- [ ] do the things above

# Dependencies

### LINKED ISSUE #347: Replace Jest with vitest
# Summary

Recently, we upgraded our project to use EcmaScript Modules (esm). Jest really doesn't seem to play nicely with them and after spending many hours, decided it might be better to look into alternative libraries. 

After researching, Vitest seems to be a drop in replacement that literally uses the same syntax and everything. 

Reasons to replace Jest with Vitest
- Vitest works natively with ESM. (fixes our main issue)
- Vitest uses the same syntax as Jest plus more!
- Vitest is much faster than Jest
- Vitest has better parallelization support
- Vitest doesn't even need a config by default. 
- Vitest has native typescript support. No need for TS-Jest additional library. 

# Requirements

- [ ] Switch it out
- [ ] Remove Jest Globals
- [ ] Remove jest configuration
- [ ] Set up new Vitest config if needed
- [ ] Update tests as necessary

# Dependencies

### LINKED ISSUE #348: Centralize Test Setup and Teardown
# Summary

-Currently our tests have a `beforeAll` and `afterAll` that is duplicated for each one. (not very efficient) Let's see if we can unify logic that is consistently run for each test. 

# Requirements

- [ ] Unify the testing setup and teardown into a new `testSetup.test.ts` file
- [ ] Remove old logic from all tests
- [ ] Ensure dbInstance creation has no issues being used as a singleton
- [ ] Clear out test db AFTER all tests are run, not after each. (this was causing issues)
- [ ] Set up facilityIds to be exported and available across all tests. 

# Dependencies

### LINKED ISSUE #349: Rework Seed Data to use Faker-Js
# Summary

- We want to use dynamic test data. Ideally, we can create as many test data objects as we need and they should all be unique. To achieve this, a generator is needed. Let's use [faker](https://fakerjs.dev/) for this. 

# Requirements

- [ ] Rework our seed data to be a factory style where it can generate n number of fake data objects.
- [ ] Integrate faker-js. 
- [ ] Upgrade our fake data to use `create` and `update` input types to match the gql schema
- [ ] Update facility fake data
- [ ] Update healthcareprofessional fake data
- [ ] Create submissions fake data (new!🥇)

# Dependencies

ermish:
# What changed

- [x] Reviewed the whole gql schema
    - [x] removed `isDeleted`
    - [x] Removed unused queries for specialties, degrees, and locales. 
    - [x] added mutations for delete facility, professional, and submission
    - [x] Fixed a bug with optional arrays that should have been required
    - [x] Removed unnecessary nested entity input types (Ex. ContactInput, DegreeInput, etc.)

### Resolves #319, Resolves #320, Resolves #321
- [x] Created and reviewed add/update inputs for each entity. Ex. `addSubmissionInput` They should now only have appropriate fields for each action (add or update)
- [x] Updated all the resolvers and service functions to use updated schema.
    - [x] FacilityService
    - [x] HealthcareProfessionalService
    - [x] SubmissionService

### Resolves #322, Resolves #323
- [x] FIXED ALL THE TESTS! 🥳 
    - [x] Updated to use new input types.

### Resolves #278
- [x] Moved validation for `updateSubmission` from resolver into the submission service 

### Resolves #345
- [x] Simplified dbSchema and fixed it to match with gqlSchema (took a lot of work)

### Resolves #303
- [x] Created `updateHealthcareProfessional` 🥳 
    - [x] Create all the validation! 💯 

### Resolves #346
- [x] Reworked `createHealthcareProfessional`
    - [x] Also created validation for `createHealthcareProfessional` 💯 

### Resolves #347, Resolves #314
- [x] 💯 Replaced Jest with Vitest! 💯   Jest was causing many problems around ESM support, parallelization issues, and not giving clear error messages when tests were failing. Vitest uses the same exact naming as Jest so there was no tests migration really needed outside of the configurations. So far, it's significantly faster and more intuitive!
- [x] Upgraded our eslint, typescript, and node to use ESM (es modules). This caused a lot of issues and now everything is up to date. 

### Resolves #348
- [x] Centralized test setup and teadown in new `testSetup.test.ts` file. This fixed a parallelization problem with tests clearing other test data.  

### Resolves #349
- [x] Fixed our seed data! It is now all generated dynamically using the faker-js library (which is awesome)
    - [x]  Updated all the tests to use this as the input now instead of a local object. 

### General improvements
- [x] Added comments everywhere! 🗨️ 😄
- [x] Simplified the spoken language to just an enum (`languageCode_iso639_3`) instead of an object. The locale name is also updated to use the same enum. 
- [x] Unified all the `add` and `create` functions to use `create` naming
- [x] Added missing logic to `updateFacility` around managing sub-entities. 
- [x] Reworked `searchSubmissions` to use new `filters` input ✨ 
- [x] Converted CustomErrors in `healthcareProfessionalService` functions to use new `Result` object and proper validations ✨

### Bug fixes
- [x] added missing facility/healthcareprofessionals from submission (had to update everything to support this) 

# What is NOT included

- [ ] Created new tests
    - [ ] Created new test to check the facility/professional id associations work
    - [ ] Created new test to validate partial field updates work for Facility and healthcare provider. (that it will change only the fields sent and won't delete the ones that were not sent in)
    - [ ] Created `approveSubmission` tests 

- [ ] Created new logic for facility and healthcareproviders to keep their "foreign key" ids in sync when either is created or updated. (for example, a new healthcareprovider is created with a facilityid included, the facility should have its `healthcareProfessionalIds` updated with the new healthcareprovider id)

- Validation needs more love across all the services. Needs a separate ticket per service. 
- `searchHealthcareProfessionals` function needs to be created.
- a lot of our mapping function probably aren't needed, can be improved, or simplified to being just a validation function

NOTE TO SELF: let's move the language object to the front end and the key will be the language code enum.
`input SpokenLanguageInput {
  languageCode_iso639_3: languageCode_iso639_3!
  nameJa: String!
  nameEn: String!
  nameNative: String!
}
`

# Testing instructions

- Ensure the app builds and runs. `yarn dev`
- Run a test query like this [one](http://localhost:4000/?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABAIq6EBiAhlAJYA2tKtCAzgBQAkAZgyvq3RFqdRigIBlBFTxQAFhT4CAhAEoiwADpIiRbjQZMWHXvX55BRHkovqtO3UVphtjokiqIAokleOPiABSVH66ctJmclAyCAAKeBDcbKy0EB70AJJgrKFEUGkoNCgaubpUYGB4ySUObmUVVaysADK0SAgAjD6ljuWVya3tAEzdtXV9jS1tncE9unTis2NuCwSjdUQAvj0IcFQMPQDmEBCH9AgAslQADqwAqnj0PddyaQg9AO4IAEYp-LnbByAwEgAA0IAAbjJaFRvudWBgQPZdJoQKZzDkMDU3KjGHAmKihB1QblUYluKwEChCURiaSQAEEN0saiAFonVF%2BYFgkDhcoCRGckA81hQPC0a5oTAgTZAA)
- Ensure all tests run and pass `yarn test`


src/services/facilityService.ts
```diff
@@ -109,118 +111,137 @@ export async function searchFacilities(filters: gqlTypes.FacilitySearchFilters =
...
+    return createFacilityResult
 }
 
 /**
- * Updates a Facility in the database with the params.
+ * Updates a Facility in the database with the params in the database based on the id. 
+ * - It will only update the fields that are provided and are not null.
+ * - If you want to create a new HealthcareProfessional, you need to call the `createHealthcareProfessional` function separately. This prevents hidden side effects.
+ * - If you want to link an existing HealthcareProfessional to a Facility, add the healthcareprofessionalId to the `healthcareProfessionalIds` array. 
+     Use the action to add or remove the association. If an id isn't in the list, no change will occur.
```
theyokohamalife, line 152: Look at you documenting things 😎

ermish: 😸

__tests__/submissions.test.ts
```diff
@@ -1,581 +1,452 @@
...
-        resolvers
-    })
+        //should not have errors
+        expect(response.body.errors).toBeUndefined()
 
-    beforeAll(async () => {
-        ({ url } = await startStandaloneServer(server, {listen: { port: 0 }}))
-        await initiatilizeFirebaseInstance()
-    })
+        const originalValues = createSubmissionRequest.variables.input
```
theyokohamalife, line 24: ♻️ same here

eslint.config.js
```diff
@@ -0,0 +1,204 @@
...
+    //         // ...jsonPlugin.configs['recommended-with-json'].rules,
+    //         // 'json/*': [
+    //         //     'error',
+    //         //     {
+    //         //         allowComments: true,
+    //         //         allowTrailingCommas: false,
+    //         //         allowTemplateLiterals: true,
+    //         //         duplicateKey: false
+    //         //     }
+    //         // ]
```
theyokohamalife, line 89: Not quite sure what's going on with this eslint config and the code that's commented out twice. Maybe you meant to remove it?

firebase.json
theyokohamalife, line 1: Is anything actually being changed in this file?

__tests__/healthcareProfessional.test.ts
```diff
@@ -1,219 +1,136 @@
...
-            facilityIds: facilityIds
-        }
-    }
-}
+import { expect, describe, it } from 'vitest'
+import { Error, ErrorCode } from '../src/result.js'
+import { generateRandomCreateHealthcareProfessionalInput } from '../src/fakeData/fakeHealthcareProfessionals.js'
+import { gqlMutation, gqlRequest } from '../utils/gqlTool.js'
+import { CreateHealthcareProfessionalInput, HealthcareProfessional } from '../src/typeDefs/gqlTypes.js'
+import { gqlApiUrl, sharedFacilityIds } from './testSetup.test.js'
```
theyokohamalife, line 7: Nice work getting the setup and teardown to work in a separate file!

src/fakeData/fakeFacilities.ts
```diff
@@ -0,0 +1,43 @@
+import * as gqlTypes from '../typeDefs/gqlTypes.js'
+import { faker, fakerJA } from '@faker-js/faker'
+
+export function generateRandomCreateFacilityInput(healthcareProfessionalIds?: string[])
+    : gqlTypes.CreateFacilityInput {
+    const sex = faker.person.sexType()
```
theyokohamalife, line 6: We haven't discussed adding `sex` to the database. I'm assuming you added this to show whether the healthcare professional is a man, woman, etc. If that's the case, I think we should use `gender` instead.

src/fakeData/fakeFacilities.ts
```diff
@@ -0,0 +1,43 @@
...
+    const firstName = faker.person.firstName()
+    const lastName = faker.person.lastName()
+    const fullEnglishName = faker.person.fullName({ firstName, lastName, sex })
```
theyokohamalife, line 9: Is there a reason the `sex` is being included in the name? This seems like something that should be a completely separate field.

src/firebaseDb.ts
```diff
@@ -13,7 +13,7 @@ export let dbInstance: Firestore
 const testFirestoreIsInitialized = async (newDbInstance: Firestore) => {
     try {
         const ref = newDbInstance.collection('facilities')
-        //eslint-disable-next-line @typescript-eslint/no-unused-vars
+        //eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
         const dbResult = await ref.limit(1).get()
```
theyokohamalife, line 17: Why not remove  the `dbResult` variable since it's not being used?

src/resolvers.ts
```diff
@@ -1,163 +1,109 @@
...
-import * as facilityService from './services/facilityService'
-import * as healthcareProfessionalService from './services/healthcareProfessionalService'
-import * as gqlType from './typeDefs/gqlTypes'
-import * as submissionService from './services/submissionService'
-import { CustomErrors, Result } from './result'
+import * as facilityService from './services/facilityService.js'
+import * as healthcareProfessionalService from './services/healthcareProfessionalService.js'
+import * as gqlType from './typeDefs/gqlTypes.js'
+import * as submissionService from './services/submissionService.js'
+import { Result } from './result.js'
```
theyokohamalife, line 6: Why are these `.ts` files being imported as `.js`? 🤔

src/resolvers.ts
```diff
@@ -1,163 +1,109 @@
...
+            convertErrorsToGqlErrors(matchingHealthcareProfessionalResult)
+            return matchingHealthcareProfessionalResult.data
+        },
         // healthcareProfessionals: async () => {
         //     const matchingProfessionals = await healthcareProfessional.searchHealthcareProfessionals(['1'])
-
+        
+        //     convertErrorsToGqlErrors(queryResults)
         //     return matchingProfessionals
         // },
```
theyokohamalife, line 37: I think we can remove this

src/services/facilityService.ts
```diff
@@ -109,125 +127,177 @@ export async function searchFacilities(filters: gqlTypes.FacilitySearchFilters =
...
+
+        // for (const currentKey in fieldsToUpdate) {
+        //     if (currentKey && currentKey in updatedDbFacility) {
+        //         updatedDbFacility[currentKey as keyof gqlTypes.UpdateFacilityInput] =
+        //             fieldsToUpdate[currentKey as keyof gqlTypes.UpdateFacilityInput]
+        //     }
+
+        //     if(currentKey === 'healthcareProfessionalIds' && fieldsToUpdate.healthcareProfessionalIds) {
+        //         fieldsToUpdate.healthcareProfessionalIds.forEach((relationship, index) => {
+        //             if(relationship.action === gqlTypes.AssociationAction.Add) {
```
theyokohamalife, line 223: It would be better to create issues for this than to check in large blocks of commented out code.

src/services/healthcareProfessionalService.ts
```diff
@@ -1,132 +1,252 @@
...
+        //business rule: createdDate cannot be set by the user.
         createdDate: new Date().toISOString(),
+        //business rule: updatedDate is updated on every change.
         updatedDate: new Date().toISOString()
-    } as dbSchema.HealthcareProfessional
+    } satisfies dbSchema.HealthcareProfessional
 }
 
-function mapDbEntityTogqlEntity(dbEntity : firebase.DocumentData) {
+//TODO solve this later
```
theyokohamalife, line 181: Please create an issue for this and remove the commented out code to it for now.

theyokohamalife (changes requested):

theyokohamalife:
It's clear that you put a ton of hard work into this and I think it's almost there!

eslint.config.js
ermish: they are technically different. I want to get them working and not lose my progress, so I kept it in. (less important right now, but it's important to protect us for future schema changes and json errors.

firebase.json
ermish: the whole file was actually reworked. I converted our eslint config to [flat config style](https://eslint.org/blog/2022/08/new-config-system-part-2/)

src/fakeData/fakeFacilities.ts
ermish: Agreed. No plans to add it at the moment, but gender would be appropriate. This is just the fake data interface so the data generated is consistent.

src/fakeData/fakeFacilities.ts
ermish: It how the fake data generator can use a consistent gendered name across fields.

src/firebaseDb.ts
ermish: It is being used to verify we can access the database. I guess we can remove the left side of the equation though 🤔

src/resolvers.ts
ermish: Tl;dr; cause the latest compilers don't yet support it as .ts 🤷‍♂️ it's the best approach to this date.

src/firebaseDb.ts
theyokohamalife: Yeah, I don't think we need to declare the variable here if it's not being used anywhere else. Could just do the `await ref` without it.

src/services/facilityService.ts
ermish: Yeah, I commented it out since I couldn't get it working in this PR but it will be in the next one. I can copy it out and add it separately

__tests__/healthcareProfessional.test.ts
```diff
@@ -1,219 +1,136 @@
...
-        expect(newHealthcareProfessionalData.degrees).toEqual(inputData.degrees)
-        expect(newHealthcareProfessionalData.spokenLanguages).toEqual(inputData.spokenLanguages)
-        expect(newHealthcareProfessionalData.acceptedInsurance).toEqual(inputData.acceptedInsurance)
-        expect(newHealthcareProfessionalData.isDeleted).toBe(false)
-        expect(newHealthcareProfessionalData.createdDate).toBeDefined()
-        expect(newHealthcareProfessionalData.updatedDate).toBeDefined()
+        expect(searchResult.body?.errors).toBeUndefined()
+
+        const searchedProfessional = searchResult.body.data.healthcareProfessional as HealthcareProfessional
+        const originalValues = createHealthcareProfessionalMutationRequest.variables.input
```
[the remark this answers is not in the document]
ermish: 💯

__tests__/submissions.test.ts
ermish: 💯

theyokohamalife (approved):

theyokohamalife:
LGTM! Thank you for all the amazing work!

PR #326, opened 2023-10-19
chore: revert renovate bot back to previous confidence level

theyokohamalife:
# What changed
This PR reverses the change that I to set the confidence level to high for the renovate bot updates. It hasn't opened any PRs to update the packages since then, and I suspect that the setting was either incorrectly implemented, or was set too high.  😅


Anissa3005 (approved):
LGTM 🚀

PR #363, opened 2023-12-01
Build out validations for specialties, insurance and spoken languages  hp

### LINKED ISSUE #362: Build out validations for specialties, insurance and spokenLanguages in healthcareProfessionalService
# Summary

# Requirements

- [x] Add more validations for specialties in HP
  - [x] Check the length of specialties name
  - [x] Check if language used matches locale
  - [x] Check if specialties contains invalid characters
 - [ ] Add more validations for insurance in HP
   - [ ] Check the insurances is  not empty
   
# Dependencies

Anissa3005:
Resolves #362 

# What changed

1. Move the remaining validations logic to the validationsHealthcareProfessional files.
2. Build out validations for specialties and minor changes to insurance and spokenLanguages validations since they are Enums.
3. Add testing for the remaining validations

# To do
Some tests are missing (if send array is to long should throw error), and need to be added later.


src/validation/validationHealthcareProfessional.ts
```diff
@@ -217,7 +217,7 @@ export function validateDegrees(
         return
     }
 
-    if (degrees && degrees.length > 64) {
+    if (degrees && degrees.length > 16) {
```
Anissa3005, line 220: The 16 logic comes from how sensei Philip has implemented  when I started working on it, so I kept that for all validations. (Try to googled it but couldn't find mush info, probably because I don't use the right keywords)

yarn.lock
theyokohamalife, line 1: You shouldn't need to make changes to the yarn.lock file since you aren't making any changes to the dependencies. If you check out the file from `main` it should remove any changes and this commit could be removed.

Anissa3005: I already fixed that mistake normally, If i check the files changed the lock file is not there anymore. I indeed checkout main yarn.lock and reverse my mistake.

Anissa3005: I have some troubles with this branch, after I rebased I needed to run yarn install this kept changing the yarn lockfile.

ermish (approved):
looks great! nice work @Anissa3005

ermish:
@Anissa3005 marking as draft since we made a bunch of changes since the last review.

theyokohamalife:
@Anissa3005 Is this still in progress or is it okay to close?

PR #371, opened 2023-12-03
chore: change yarn version

theyokohamalife:
Currently, we're experiencing an error during the build process on Digital Ocean that the current app stack doesn't support Yarn 2+.

This change will set our Yarn to `1.22.21`. It will also use`tsx` instead of `ts-node`, because it seems to work out of the box with the Yarn downgrade.


package.json
```diff
@@ -35,13 +35,15 @@
...
+    "packageManager": "yarn@1.22.21",
     "engines": {
-        "node": "18.x",
-        "yarn": "^4.0.1"
+        "node": "18.18.2",
+        "yarn": "1.22.21"
     },
     "devDependencies": {
         "@faker-js/faker": "^8.2.0",
+        "@firebase/app": "^0.9.24",
```
ermish, line 45: did we need to add these?

tsconfig.json
```diff
@@ -14,9 +14,6 @@
         "esModuleInterop": true,
         "skipLibCheck": true
     },
-    "ts-node": {
```
ermish, line 17: this will have a big negative hit to performance

PR #373, opened 2023-12-04
refactor: add -a option to copyfiles

theyokohamalife:
Change `copyfiles` to `copyfiles -a` to include files that start with a dot.


ermish (approved):
🚢 eet

PR #376, opened 2023-12-04
refactore: remove single quotes from copyfiles

theyokohamalife:
Removes the single quotes to the copyfiles cmd since [the docs](https://www.npmjs.com/package/copyfiles) shows them as not needed.


ermish (approved):
🚢 it

PR #377, opened 2023-12-04
chore: add custom script for digital ocean

theyokohamalife:
Adds a custom build script for Digital Ocean. According to the [docs](https://docs.digitalocean.com/products/app-platform/reference/buildpacks/nodejs/):

> App Platform doesn’t install devDependencies by default during the build process. If your build command requires devDependencies, configure the following custom script. This script installs devDependencies, runs your build command, and removes any installed devDependencies before deployment.


package.json
```diff
@@ -4,6 +4,7 @@
     "description": "Backend for Find a Doc, Japan",
     "main": "index.js",
     "scripts": {
+        "build:digitalocean": "yarn install --production=false && yarn prod:build && rm -rf node_modules && yarn install --production --frozen-lockfile",
```
ermish, line 7: nit: should call it prod:builddigitalocean

ermish (approved):
☘️

ermish (changes requested):
we already have  a prod build `yarn prod:build`. let's add to that script

PR #383, opened 2023-12-04
refactor: add no-daemon flag to pm2

theyokohamalife:
Adds the `--no-daemon` flag to pm2 so it runs in the foreground.


Anissa3005 (approved):
LGTM 😽

PR #385, opened 2023-12-04
chore: change prod port to 8080

theyokohamalife:
Change prod port from 4000 to 8080.


ermish (approved):
ports are clear! ship it

PR #393, opened 2023-12-06
migrated to npm...for now

ermish:
# What changed
migrated from yarn to npm.....because digitial ocean doesn't support yarn past v1. :(


Anissa3005 (approved):
🐈

PR #410, opened 2023-12-12
fix: point the prod env to correct urls

theyokohamalife:
Updates the URLS for the API and Website in the production environment file.

### DIFF
--- a/.env.prod
+++ b/.env.prod
@@ -1,8 +1,8 @@
 TEST_ENABLED = false
 SERVER_HOST = "0.0.0.0"
 SERVER_PORT = 4000
-API_URL = "https://findadoc.jp/api"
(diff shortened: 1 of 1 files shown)


Anissa3005 (approved):
LGTM 🚀

.env.prod
```diff
@@ -1,8 +1,8 @@
 TEST_ENABLED = false
 SERVER_HOST = "0.0.0.0"
 SERVER_PORT = 4000
-API_URL = "https://findadoc.jp/api"
-WEBSITE_URL = "https://findadoc.jp"
+API_URL = "https://api.findadoc.jp/api"
```
ermish, line 4: we should drop the /api part. We'll need to update the website at the same time

PR #417, opened 2023-12-14
Build out validations for healthcare professional

### LINKED ISSUE #362: Build out validations for specialties, insurance and spokenLanguages in healthcareProfessionalService

Anissa3005:
Resolves #362 

# What changed

1. Move the remaining validations logic to the validationsHealthcareProfessional files.
2. Build out validations minor changes to specialties, insurance, spokenLanguages validations since they are Enums.
3. Add testing for the remaining validations

# Testing instructions

Some tests are still missing (if send array is to long should throw error), and need to be added later.

# To do

Since there been a lot of changes over the past week, some validations are now incomplete or should be approach different.


ermish (approved):
LGTM!

PR #434, opened 2024-01-30
Migrate formatting to Eslint Stylistic

### LINKED ISSUE #353: Migrate eslint rules to use `@stylistic`
# Summary

eslint is splitting out their rules into a separate project called `[stylistic](https://eslint.style/guide/why)`. There's a lot of nice reasons for this. For us it would: 
1) simplify our rules between js rules, typescript rules, vue rules to where we don't have to use overrides. one rule works for all files. 
2) simplify package dependencies to have all rules come from an official, single source. 

# Requirements

- [ ] Update our eslint to use the new rules 
- [ ] Follow the [migration guide](https://eslint.style/guide/migration)

# Dependencies

- None

Anissa3005:
Resolves #353 

# What changed

- I have added the ESLint stylistic plugin to be able to keep using our formatting rules in  ESLint config file.
- I have change the formatting rules that will be deprecated. The links below lists all the rules that will be effected.

These changes triggered formatting in a few files, so there are quite few files that have formatting changes in them. Please check if these changes are valid.

These are all the rules that are covered by ESLint Stylistic: https://eslint.style/packages/default
ESLint Deprecation of Formatting rules: https://eslint.org/blog/2023/10/deprecating-formatting-rules/

# Testing instructions
NA


eslint.config.js
```diff
@@ -1,15 +1,16 @@
...
-import eslintJsPlugin from '@eslint/js'
+import globals from "globals";
+import ts from "@typescript-eslint/eslint-plugin";
+import tsParser from "@typescript-eslint/parser";
+import * as graphqlESLint from "@graphql-eslint/eslint-plugin";
+import eslintJsPlugin from "@eslint/js";
+import stylistic from "@stylistic/eslint-plugin";
 
-const gqlSchemaPath = './src/typeDefs/schema.graphql'
+const gqlSchemaPath = "./src/typeDefs/schema.graphql";
```
Anissa3005, line 8: My Prettier keeps changing the single quotes to double in certain files, running ESLint --fix does nor resolve this issue. So that made me think if we need to have a Prettier config file or have a .prettierignore and use the ESLint --fix only (I would go with the last one?)

eslint.config.js
```diff
@@ -102,113 +100,126 @@ export default [
...
-            'comma-dangle': ['error', 'never'],
-            'comma-spacing': ['error', { before: false, after: true }],
-            'comma-style': 'error',
-            'computed-property-spacing': ['error', 'never'],
-            'function-paren-newline': ['error', 'consistent'],
-            indent: [
-                'error',
+            // arrow-body-style is not supported by @stylistic, TypeError said to use '@/arrow-body-style', so for now I have added it like that
+            "@/arrow-body-style": "error",
+            "@stylistic/quotes": ["error", "single"],
```
Anissa3005, line 143: I have added this line, but not sure if this was necessary.

eslint.config.js
```diff
@@ -1,15 +1,16 @@
-import globals from 'globals'
```
ermish, line 1: your linter for some reason is adding semicolons and then adding dangling commas 🤔

eslint.config.js
ermish: ohhhhh.....we don't use prettier. that's why it's conflicting.

eslint.config.js
ermish: it should replace the quotes line further down 👍

eslint.config.js
Anissa3005: So I have Prettier as a plugin for my VS Code, the thing I have read online is to add a prettierignore file to your project? Or is this not correct? 

Because I remember that LaShawn had also something similar with adding double quotes.

Anissa3005: Ah Sorry Philip, I just had to disable the plugin for the find a doc workspace!

ermish (approved):
LGTM! nice work. 
I see some dangling commas still, but no reason to block

ermish:
merge when you're ready @Anissa3005

PR #443, opened 2024-04-08
create table for audit logs

Anissa3005:
We want to keep track of all the updates and deletions in case we need to rollback

Resolves #[Issue number]

# What changed

# Testing instructions


src/typeDefs/dbSchema.ts
```diff
@@ -26,6 +26,8 @@ export type Submission = Omit<gqlType.Submission, 'facility' | 'healthcareProfes
     healthcareProfessionals: gqlType.HealthcareProfessionalSubmission[]
 }
 
+export type AuditLogs = Omit<gqlType.AuditLogs, 'AuditLogs'>
+
```
Anissa3005, line 30: @ermish Can you briefly explain what this file exactly does? Is it the types that are stores in the DB? Not sure if I fully understand this part.

src/typeDefs/schema.graphql
```diff
@@ -157,6 +158,35 @@ input SubmissionSearchFilters {
   offset: Int
 }
 
+type AuditLogs {
+  id: ID!
+  actionType: ActionType!
+  tableName: TableName!
+  schemaVersion: Float!
+  jsonData: String!
```
Anissa3005, line 166: @ermish my research lead me to using `String!` for JSON data since there is no scalar type yet for json data (or we have to created it ourself?)

ermish: I think that's right

src/typeDefs/dbSchema.ts
ermish: I added some comments around how the file works, but I guess I didn't explain what it was 😓 
these types are meant to match the data that is saved in the database.

It's not guaranteed that the data object structure is identical in front-end, api calls, backend logic/services, and in the database. for example, you don't want security data in your api objects, but you do in your service logic. Similarly, the way you save your data might be different (normalized vs denormalized for example) than how you use the data in your service. 

The Omit<> type is a shortcut instead of making a whole new type object, you can  just remove certain properties from an existing one.

src/typeDefs/schema.graphql
```diff
@@ -159,6 +160,41 @@ input SubmissionSearchFilters {
   offset: Int
 }
 
+type AuditLogs {
+  id: ID!
+  actionType: ActionType!
+  tableName: TableName!
```
ermish, line 166: I  realized this is a mistake on my requirements I gave here. We shouldn't expose our tables to the public. (security vulnerability)
Let's change this to `entity` or `model`, `objectType` or something like that. the enum values can still be reused

ermish (changes requested):

ermish (approved):
time for auditing!

PR #454, opened 2024-05-27
chore: update readMe to be more comprehensive

NabbeunNabi:
No issue linked due to no view for backend

# What changed
The readMe was a bit outdated and therefore I could see new people coming in trying to run the dummy data in the local database have trouble setting it up. I went through and updated the language that was there and added instructions for two parts of Java that need to be installed. I also commented out the husky command. There is no script for it. Whether there should be or not it would just be a stumbling block for people who do not know why they are getting the error or might now be familiar with how we are implementing husky and linting

# Testing instructions

You can follow step by step to see if it makes sense to you

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,33 +72,29 @@ For security, we run a local database so we don't break production!
 npm -g i firebase-tools
 ```
 
+**To be able to run the firestore emulator you need to have Java version 11 or later installed. If not installed, please follow the next step (a.2).**
+
 #### a.2 Install Java
 
 on a mac, we recommend using [homebrew](https://brew.sh/)
 
 ```sh
-brew install java
+brew install openjdk
 ```
 
 on a windows pc, we recommend using [chocolatey](https://chocolatey.org/)
 
 ```sh
-choco install java
+choco install openjdk
 ```
+
+**You might have to restart vs code for step a.2 and a.3 all of these to take effect in the project**
 
 #### a.3 Log into Firebase CLI:
 
 ```sh
 firebase login
 ```
-
-#### a.4 Set the CLI to use the project:
-
-```sh
-firebase use --add
-```
-
-Select "Use an existing project"
 
 ### 3.b Running the database locally
 
@@ -130,7 +126,7 @@ To set up the Firebase Service Account for this project, follow these steps:
 ## 4. Run the API
 
 ```sh
-npm run dev
+npm run dev:startserver
 ```
 
 That's it!
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
         "test": "echo '🧶 Installing npm dependencies...🧶' && npm install && vitest",
         "test:dockerstart": "docker-compose -f 'docker/docker-compose-test.yml' up -d --build",
         "test:dockerstop": "docker-compose -f 'docker/docker-compose-test.yml' down",
+        "prepare": "husky install",
         "prod": "npm run prod:dockerstop && npm run prod:dockerstart",
         "prod:build": "echo '🧶 Installing npm dependencies...🧶' && npm install --immutable --immutable-cache && echo '🧹 cleaning dist folder...' && npx rimraf dist/ && echo '➡ ➡ copying files...' && npx copyfiles -a -V src/typeDefs/schema.graphql .env.prod dist && echo '💻✨ compiling code..' && tsc && echo '💻✨ successfully built!'",
         "prod:startserver": "echo '💻 💻 💻✨ Starting prod server...' && cross-env NODE_ENV=production PIDUSAGE_USE_PS=true pm2 start dist/src/index.js --no-daemon && pm2 logs --lines 100",


README.md
```diff
@@ -130,7 +154,7 @@ To set up the Firebase Service Account for this project, follow these steps:
 ## 4. Run the API
 
 ```sh
-npm run dev
+npm run dev:startserver
```
Anissa3005, line 157: This needs to be `npm run dev` since that sill install, create typedefs, seed and start the server. The `npm run dev:startserver` is called in the `npm run dev`

README.md
```diff
@@ -48,57 +48,81 @@ We love and welcome contributions to our front-end repository which can be found
...
 ```sh
-brew install java
+brew install openjdk
 ```
 
 on a windows pc, we recommend using [chocolatey](https://chocolatey.org/)
 
 ```sh
-choco install java
+choco install openjdk
```
Anissa3005, line 88: If you see the documentation of the firestore local emulator, they mention to install Java JDK. So I'm not sure if it's a good idea to go for `openJDK` . 

So in this part I agree that we should specify the the version of Java that needs to be installed so we need don't need the upgrading part in the next part.

I think it's better to stick with the firestore local emulator documentation: https://firebase.google.com/docs/emulator-suite/install_and_configure

README.md
```diff
@@ -48,57 +48,81 @@ We love and welcome contributions to our front-end repository which can be found
...
+### 2.a Setup Firebase CLI
 
 #### a.1 Install the Firebase CLI if you do not have it installed on your machine:
 
 ```sh
 npm -g i firebase-tools
 ```
 
-#### a.2 Install Java
+#### a.2 Install Java to the path
```
Anissa3005, line 77: Keep this install Java, since you need to have Java on your local machine to run the emulator

README.md
```diff
@@ -48,57 +48,81 @@ We love and welcome contributions to our front-end repository which can be found
...
+You should see an updated version of Java like below
+
+```sh
+java version "22.0.1" 2024-04-16
+```
+
+#### You might have to restart vs code for step a.2 and a.3 all of these to take effect in the project
+
+#### a.3 Log into Firebase CLI:
+
```
Anissa3005, line 122: The `firebase use --add` need to be in the readme in case people have other projects running with firebase.(this was the case for me), so you need to be able to point to the correct project.

https://firebase.google.com/docs/cli

Anissa3005 (changes requested):

README.md
NabbeunNabi: It did not work when I did that I would get this error
```> findadoc-server@1.0.0 generate
> echo '➡ ➡ ➡ Generating types from graphql schema ➡ ➡ ➡' && graphql-codegen --config ./typesgeneratorconfig.ts

'➡ ➡ ➡ Generating types from graphql schema ➡ ➡ ➡' 
(node:15732) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)✔ Parse Configuration
⚠ Generate outputs
  ❯ Generate to src/typeDefs/gqlTypes.ts
    ✔ Load GraphQL schemas
    ✔ Load GraphQL documents
    ✖ Cannot use GraphQLObjectType "Contact" from another module or rea…  
Running lifecycle hook "afterStart" scripts...
[CLI] Loading Schemas
[CLI] Loading Documents
[CLI] Generating output
```

README.md
NabbeunNabi: So just delete this altogether. Because running choco install java gives this error

```Chocolatey v2.2.2
Installing the following packages:
java
By installing, you accept licenses for the packages.
java not installed. The package was not found with the source(s) listed.
 Source(s): 'https://community.chocolatey.org/api/v2/'
 NOTE: When you specify explicit sources, it overrides default sources.
If the package version is a prerelease and you didn't specify `--pre`,
 the package may not be found.
Please see https://docs.chocolatey.org/en-us/troubleshooting for more
 assistance.

Chocolatey installed 0/1 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).

Failures
 - java - java not installed. The package was not found with the source(s) listed.
 Source(s): 'https://community.chocolatey.org/api/v2/'
 NOTE: When you specify explicit sources, it overrides default sources.
If the package version is a prerelease and you didn't specify `--pre`,
 the package may not be found.
Please see https://docs.chocolatey.org/en-us/troubleshooting for more
 assistance.


```

README.md
NabbeunNabi: Then this needs to be given conditionals if you have other projects. Technically I do not belong to the firebase for findadoc. So this is what I saw when I used that command 

```
$ firebase use --add
(node:6176) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)? Which project do you want to add? (Use arrow keys)
```

And if I use the arrow keys I get this.

`? Which project do you want to add?`

There are no options. So I could try this forever and keep getting this error
`Error: An unexpected error has occurred.`

So we should explain if you are part of the firebase you can select it otherwise you can skip this step? What do you think?

README.md
```diff
@@ -77,28 +79,50 @@ npm -g i firebase-tools
 on a mac, we recommend using [homebrew](https://brew.sh/)
 
 ```sh
-brew install java
+brew install openjdk
```
theyokohamalife, line 82: Was this tested on macOS? I wouldn't recommend changing it if not.

NabbeunNabi: Cannot test it on mac. Need people on other OS to test the steps as well 😥

README.md
theyokohamalife: @Anissa3005 @NabbeunNabi I think that this level of detail is something we should exclude from the README altogether. The steps for installing Firebase CLI will change over time and we don't want to have to keep updating our README to reflect those changes. Instead, we should just tell users to follow the steps in the official documentation and link them to it.

theyokohamalife:
@Anissa3005 @NabbeunNabi I pushed a small commit to reintroduce Husky to the README. Could you both confirm that it works? If so, we can update the PR title and description.

README.md
NabbeunNabi: We figured it out!

README.md
NabbeunNabi: @Anissa3005 mentioned that she and Evan put this here because the Firebase CLI doesn't mention the steps for installing Java and kind of assumes you have it. So we thought it might be better to make it less but also keep it for those who are unfamiliar at all with Java. Anissa also said the chocolately command I linked is the one she used and the previous command was incorrect

README.md
NabbeunNabi: Fixed

README.md
NabbeunNabi: As mentioned. The previous command java was incorrect. Should have always been `openjdk`

theyokohamalife (approved):
LGTM ✨

theyokohamalife:
@Anissa3005 You'll need to approve this since you requested changes.

Anissa3005 (approved):
LGTM

Issue #467, opened 2024-06-04
Add fake submissions data to the seeding file.


Anissa3005:
# Summary

We didn't had our fake submissions data in our seeding file, therefore we had no testing data for the frontend.

# Requirements

- [ ] Add the fake submissions data to the seeding file

# Dependencies

- None (Add any issue dependencies here)

PR #468, opened 2024-06-04
Add fake submissions data to seeding file

Anissa3005:
Resolves: #467 

The the fake submissions data was missing from the seeding file, therefore we did not have any seeded data for submissions

# What changed

- Add the fake data to the seeding file

# Testing instructions
Run `npm run dev:startlocaldb`
Then `npm run dev`

Visit emulator database and see the seeded data for submissions. 

# To do

This is a quick fix, so we can work on the frontend. This means I didn't check in depth that fake data is as it should be.


NabbeunNabi:
Thanks so much for working on this!

ermish (approved):
LGTM!

Issue #482, opened 2024-06-16
Add map latitude and longitude as types for Facilities query


NabbeunNabi:
# Summary
Right now our database stores these in the database. But they are not generated as types for the frontend by the backend.

# Requirements

- [ ] Add the types

# Dependencies

- None

NabbeunNabi:
apparently it is. Don't know why it isn't autoshowing on a branch I was on. My mistake

NabbeunNabi:
Never mind it wasn't part of the Facility submission but if the submission is updated we would want it as part of it.

PR #485, opened 2024-06-17
chore: add coordinates to schema `FacilitySubmission`

NabbeunNabi:
# What changed
I added the coordinates to the types on the backend in #482 . This one adds the types to the schema which is what generates the types for the frontend. Thanks to @Anissa3005 For explaining to me that the `schema.graphql` file is what needs to be changed

# Testing instructions
Run the server with:

```
npm generate 
```
And see if the type for the floats of the coordinates is part of the `FacilitySubmission` type.


Anissa3005 (approved):
LGTM 🚀

Issue #234, opened 2023-08-27
Create a function to fetch PlaceID from Google Maps link


theyokohamalife:
Updated user flow:

The backend will have a mutation for updating the location and triggering the google search.
the front end will have a button on the moderator page to manually trigger the mutation
the front end can update the button to say something like "rerun" when the data already exists. (it can be rerun, but want to call out that it has already been run)
We don't want to auto-run because not all submissions will need it plus it will prevent costs on spam submissions

theyokohamalife:
You can find a place ID by doing a ["Place Search"](https://developers.google.com/maps/documentation/places/web-service/search-find-place) request.

Examples

Query by radius for given coordinates:
```
https://maps.googleapis.com/maps/api/place/findplacefromtext/json
  ?fields=formatted_address%2Cname%2Crating%2Copening_hours%2Cgeometry
  &input=mongolian
  &inputtype=textquery
  &locationbias=circle%3A2000%4047.6918452%2C-122.2226413
  &key=YOUR_API_KEY
```

Query by phone number:

```
https://maps.googleapis.com/maps/api/place/findplacefromtext/json
  ?input=%2B16502530000
  &inputtype=phonenumber
  &key=YOUR_API_KEY
```

Query by name of place:

```
https://maps.googleapis.com/maps/api/place/findplacefromtext/json
  ?fields=formatted_address%2Cname%2Crating%2Copening_hours%2Cgeometry
  &input=Museum%20of%20Contemporary%20Art%20Australia
  &inputtype=textquery
  &key=YOUR_API_KEY
```

ermish:
https://github.com/ourjapanlife/findadoc-server/pull/498#issuecomment-2388617987
added some new requirements. Will update the ticket

PR #498, opened 2024-07-04
feat: implement function to procure facility submission details

NabbeunNabi:
Resolves issue #234 

The documentation that describes what I did is:
-[Google Places API documentation](https://developers.google.com/maps/documentation/places/web-service/text-search)
-[Pricing](https://developers.google.com/maps/documentation/places/web-service/usage-and-billing)
**Pricing is only $2.83 per 1000 calls**

## Flow
I made comments in the code to help follow. But this is the flow of the file:

1.  Use puppeteer to open  a headerless browser that will wait for the page to redirect to the desktop version of the link. This will get us a url with the coordinates.
2. From this url `regex` is used to extract the coordinates that are in every url
3. The coordinates are then used to do a google places search using a text query and location bias returning the closest relevant response to the query. This is because the coordinates should be as exact as possible. (We might need to think more if they are in the same building. Currently I am returning all the fields because we can implement our own filters later for other things if we scale like wheelchair accessibility.)
4. The prior three steps are then wrapped up in a final function that can be used

## Reasoning for flow
I tried multiple ways to get Tokyo Midtown Clinic will you will see is what I tested in the testing header. Following are the other things I tried and read about:
- Using reverse geocoding to use the coordinates to find the exact place. This wasn't very viable as it would return anything that is in the same building whether or not hospital and we shouldn't implement logic that google is already doing for us to filter. I tested it with different options and it never even returned the desired location.
- I used other parts of the places api. Again just using coordinates did not work for the previous reason plus I tried to include the parameter `types` with `hospital` in the argument array. But that seemed to prioritize hospitals only and clinics is not a viable argument
- I used Nearby Search in the documentation but it is less reliable since there are many things nearby the coordinates
- I tried to find the place from text using just the url, but it seems like the algorithm they use does not want to use it's own details.
- This had me settle on text query with the `body` that I wrote in the code included in the `POST` request

## Testing Instructions (for now)
1.  Add a `.env` with a `GOOGLE_MAPS_API_KEY`
2.  If you have `tsx` installed run:
```sh
tsx ./utils/submissionDataFromGoogleMaps.ts
```
**OR**

In the terminal run:
```sh
npx tsc utils/submissionDataFromGoogleMaps.ts
```
And then change the generated file from `.js` to `.cjs` so the next command can run the common js generations
and run: 
```sh
node utils/submissionDataFromGoogleMaps.cjs
```
**This was used to generate the following code twice for Tokyo Midtown Clinic using the google places API.**

**Try 1**
```[
  {
    name: 'places/ChIJ30G_l3iLGGARXEK75Rc-3Ww',
    id: 'ChIJ30G_l3iLGGARXEK75Rc-3Ww',
    types: [ 'doctor', 'health', 'point_of_interest', 'establishment' ],
    nationalPhoneNumber: '03-5413-0080',
    internationalPhoneNumber: '+81 3-5413-0080',
    formattedAddress: 'Japan, 〒107-6206 Tokyo, Minato City, Akasaka, 9-chōme−7−１, Midtown Tower, 6F',
    addressComponents: [
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object]
    ],
    plusCode: {
      globalCode: '8Q7XMP8J+CG',
      compoundCode: 'MP8J+CG Minato City, Tokyo, Japan'
    },
    location: { latitude: 35.6661201, longitude: 139.7312888 },
    viewport: { low: [Object], high: [Object] },
    rating: 2.7,
    googleMapsUri: 'https://maps.google.com/?cid=7844494398308696668',
    websiteUri: 'https://www.tokyomidtown-mc.jp/',
    regularOpeningHours: { openNow: true, periods: [Array], weekdayDescriptions: [Array] },
    utcOffsetMinutes: 540,
    adrFormatAddress: '<span class="country-name">Japan</span>, <span class="postal-code">〒107-6206</span> <span class="region">Tokyo</span>, <span class="street-addres
s">Minato City, Akasaka, 9-chōme−7−１, Midtown Tower</span>, 6F',
    businessStatus: 'OPERATIONAL',
    userRatingCount: 204,
    iconMaskBaseUri: 'https://maps.gstatic.com/mapfiles/place_api/icons/v2/generic_pinlet',
    iconBackgroundColor: '#7B9EB0',
    displayName: { text: 'Tokyo Midtown Clinic', languageCode: 'en' },
    currentOpeningHours: { openNow: true, periods: [Array], weekdayDescriptions: [Array] },
    shortFormattedAddress: 'Minato City, Akasaka, 9-chōme−7−1 6F Midtown Tower',
    reviews: [ [Object], [Object], [Object], [Object], [Object] ],
    photos: [
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object]
    ],
    restroom: true,
    accessibilityOptions: {
      wheelchairAccessibleParking: true,
      wheelchairAccessibleEntrance: true,
      wheelchairAccessibleRestroom: true
    }
  }
]
```

**Try 2**
```
[
  {
    name: 'places/ChIJ30G_l3iLGGARXEK75Rc-3Ww',
    id: 'ChIJ30G_l3iLGGARXEK75Rc-3Ww',
    types: [ 'doctor', 'point_of_interest', 'health', 'establishment' ],
    nationalPhoneNumber: '03-5413-0080',
    internationalPhoneNumber: '+81 3-5413-0080',
    formattedAddress: 'Japan, 〒107-6206 Tokyo, Minato City, Akasaka, 9-chōme−7−１, Midtown Tower, 6F',
    addressComponents: [
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object]
    ],
    plusCode: {
      globalCode: '8Q7XMP8J+CG',
      compoundCode: 'MP8J+CG Minato City, Tokyo, Japan'
    },
    location: { latitude: 35.6661201, longitude: 139.7312888 },
    viewport: { low: [Object], high: [Object] },
    rating: 2.7,
    googleMapsUri: 'https://maps.google.com/?cid=7844494398308696668',
    websiteUri: 'https://www.tokyomidtown-mc.jp/',
    regularOpeningHours: { openNow: true, periods: [Array], weekdayDescriptions: [Array] },
    utcOffsetMinutes: 540,
    adrFormatAddress: '<span class="country-name">Japan</span>, <span class="postal-code">〒107-6206</span> <span class="region">Tokyo</span>, <span class="street-addres
s">Minato City, Akasaka, 9-chōme−7−１, Midtown Tower</span>, 6F',
    businessStatus: 'OPERATIONAL',
    userRatingCount: 204,
    iconMaskBaseUri: 'https://maps.gstatic.com/mapfiles/place_api/icons/v2/generic_pinlet',
    iconBackgroundColor: '#7B9EB0',
    displayName: { text: 'Tokyo Midtown Clinic', languageCode: 'en' },
    currentOpeningHours: { openNow: true, periods: [Array], weekdayDescriptions: [Array] },
    shortFormattedAddress: 'Minato City, Akasaka, 9-chōme−7−1 6F Midtown Tower',
    reviews: [ [Object], [Object], [Object], [Object], [Object] ],
    photos: [
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object]
    ],
    restroom: true,
    accessibilityOptions: {
      wheelchairAccessibleParking: true,
      wheelchairAccessibleEntrance: true,
      wheelchairAccessibleRestroom: true
    }
  }
]
```

This is the frontend with the autofill data! It has the correct place. For each of these URLS

image


utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,124 @@
+import puppeteer from 'puppeteer'
+import dotenv from 'dotenv'
+
+//Load the api key for google maps
+dotenv.config()
+
```
Anissa3005, line 6: dotenv is not necessary anymore with the new node versions I think we can just use process.env.NAME_OF_ENV

utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,124 @@
...
+// This function ensures that we can redirect in a headerless browser using puppeteer to extract url once it has coordinates
+const extractRedirectedUrl = async (initialUrl: string): Promise<string> => {
+  const browser = await puppeteer.launch({ headless: true }) // ensures it doesn't open an actual browser
```
Anissa3005, line 9: Why is this step necessary? Can you add a JSDoc comment so there is a clear explanation. 

I assume it's because googlemapURLs on the phone are different and we want to get the 'long' url?

utils/submissionDataFromGoogleMaps.ts
NabbeunNabi: Oh cool. I didn't know that.

utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,124 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+      'X-Goog-Api-Key': apiKey,
+      'X-Goog-FieldMask': '*'
+    },
+    body: JSON.stringify(requestBody)
+  }
+
+  try {
+    const response = await fetch(url, fetchOptions)
```
Anissa3005, line 76: My opinion, we should not use the build in fetch() I prefer Axios over fetch(). There are better build-in behaviors

utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    if (results.length) {
+      return results
+    } else {
+      console.error('No hospitals or clinics found for the given coordinates.')
+      return null
+    }
+  } catch (error) {
+    console.error('Error fetching nearby locations:', error)
+    throw error
```
Anissa3005, line 94: This Throw is not necessary I believe. Since you are throwing your error on line 80 for example and it will catch it here. Here you should log the error and return an error (something done for the future) but the throw error should be removed for now the console.log is ok

utils/submissionDataFromGoogleMaps.ts
NabbeunNabi: Yea I can add it to the code. I have a comment at the top. Maybe it isn't clear. I wrote in the PR descriptions the thought process. But as an example. The url for a mobile url from Apple is something like https://maps.app.goo.gl/wQMavNqFCv3HVVqh7?g_st=ic

But we need to get the coordinates to use the logic below to use the Google places api in the most efficient manner with the data the user submits. 

This step ensures we get a consistent url with coordinates. Since the pattern on a browser is consistently like this

https://www.google.com/maps/place/McDonald's+Akasaka+Station/@35.6722412,139.7360999,16z/data=!3m1!4b1!4m6!3m5!1s0x60188b80c35827c9:0x70699f09e29c704!8m2!3d35.6722412!4d139.7360999!16s%2Fg%2F1tmkkzd9?entry=ttu

utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,124 @@
...
+      'X-Goog-FieldMask': '*'
+    },
+    body: JSON.stringify(requestBody)
+  }
+
+  try {
+    const response = await fetch(url, fetchOptions)
+
+    if (!response.ok) {
+      const errorText = await response.text()
```
Anissa3005, line 79: With Axios for example this part is not necessary it will catch the 404/500... errors

utils/submissionDataFromGoogleMaps.ts
NabbeunNabi: Fine with me. I just used fetch cuz @ermish doesn't usually like added libaries and dependencies

utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,124 @@
...
+    console.error('No coordinates found in the URL.')
+    return null
+  }
+}
+
+
+
+// This function uses the places api to get the closest clinic to the coordinates in the url
+const getNearbyLocationsFromCoordinates = async (latitude: string, longitude: string, apiKey: string): Promise<any | null> => {
+  const url = 'https://places.googleapis.com/v1/places:searchText'
```
Anissa3005, line 46: What I have read so far is that you can fetch the place id by using the coordinates. So this could be another possibility, fetch the place id by coordinates, than fetch the location details using the place id. Not sure if this is something you already tried or did

NabbeunNabi: The problem with that solution is that in Japan. Tall buildings can have many clinics and they are at the same coordinates

utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,124 @@
...
+  } catch (error) {
+    console.error('Error fetching nearby locations:', error)
+    throw error
+  }
+}
+
+
+// This uses all the helper functions in order to get the data we want
+const getFacilityDetailsForSubmission = async (initialUrl: string, apiKey: string): Promise<any | null> => {
+  try {
```
Anissa3005, line 101: This function I would put on the top of file, since this is the main feature and the other functions are helper functions. This would make the code also more readable

utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,124 @@
...
+      throw new Error('Could not extract coordinates.')
+    }
+  } catch (error) {
+    console.error('Error in getFacilityDetailsForSubmission:', error)
+    throw error
+  }
+}
+
+// Exporting the function to the files where it needs to be used
+export { getFacilityDetailsForSubmission }
```
Anissa3005, line 118: use export in front of the function name

utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,124 @@
...
+
+
+// This uses all the helper functions in order to get the data we want
+const getFacilityDetailsForSubmission = async (initialUrl: string, apiKey: string): Promise<any | null> => {
+  try {
+    const redirectedUrl = await extractRedirectedUrl(initialUrl)
+    const coordinates = extractCoordinatesFromRedirectedUrl(redirectedUrl)
+    if (coordinates) {
+      const addressDetails = await getNearbyLocationsFromCoordinates(coordinates.latitude, coordinates.longitude, apiKey)
+      console.log(addressDetails)
```
Anissa3005, line 106: I think the fetching logic should be in this function, I think there is no need for a separate `getNearbyLocationsFromCoordinates` function. I think we can handle this in one function

I would even name this function something more like `getGoogleMapUrlLocationDetails` since we are getting the details from the google map url.

utils/submissionDataFromGoogleMaps.ts
Anissa3005: Hmm yeah I see. Maybe LaShawn's proposal is maybe a better way than by using a phone number, since than we are 100% sure to get the write place if we assume that the user gives up the right phone number. 

How do we make sure now that we get the right place?

utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  try {
+    const response = await fetch(url, fetchOptions)
+
+    if (!response.ok) {
+      const errorText = await response.text()
+      throw new Error(`HTTP error! Status: ${response.status} - ${errorText}`)
+    }
+
+    const data = await response.json()
```
Anissa3005, line 83: This variable names should be more descriptive, same for response and result. This would make it more clear what is exactly happening.

theyokohamalife:
This looks like a really good start! My only concern is about the cost. How can we prevent being charged from fake submissions, but still pre-fill the submission fields to satisfy the submission input? Let's discuss the pros and cons with doing this on the backend if we decide to implement a button for moderators to click to fetch the place data.

utils/submissionDataFromGoogleMaps.ts
NabbeunNabi: We don't know we get the right place all the time unfortunately. That is really the moderators job to double check. This is just to help with filling the form and less work if what they confirmed is correct right away

NabbeunNabi:
> This looks like a really good start! My only concern is about the cost. How can we prevent being charged from fake submissions, but still pre-fill the submission fields to satisfy the submission input? Let's discuss the pros and cons with doing this on the backend or frontend if we decide to implement a button for moderators to click to fetch the place data.

Based on the discussions I have had with @ermish . As long as the Regex checks are working it will at least be a valid google maps url that is a location due to the redirection and the finding of the coordinates and making sure it is a google maps URL. And at this point it would be easiest to autofill on submission. If we get to the point where people start trying to submit fake URLS with bots we can implement a captcha library to prevent this.

utils/submissionDataFromGoogleMaps.ts
NabbeunNabi: I tried this. but for some reason it wasn't loading the variable and showing as undefined. Could be how I was using it but just in case I put it back for now

NabbeunNabi:
> > This looks like a really good start! My only concern is about the cost. How can we prevent being charged from fake submissions, but still pre-fill the submission fields to satisfy the submission input? Let's discuss the pros and cons with doing this on the backend or frontend if we decide to implement a button for moderators to click to fetch the place data.
> 
> Based on the discussions I have had with @ermish . As long as the Regex checks are working it will at least be a valid google maps url that is a location due to the redirection and the finding of the coordinates and making sure it is a google maps URL. And at this point it would be easiest to autofill on submission. If we get to the point where people start trying to submit fake URLS with bots we can implement a captcha library to prevent this.

I actually refactored. I think overall it would be better as a button on the moderation panel. It does take some time. So we want the submission created as quickly as possible. The moderators can then autofill the database with the added mutation.

src/resolvers.ts
```diff
@@ -212,6 +213,66 @@ const resolvers = {
             return createSubmissionResult.data
         },
 
+        moderationPanelUpdateSubmission: async (_parent: unknown, args: {
```
ermish, line 216: this name doesn't make sense to me. What is the action being performed that's different from `updateSubmission`? I would either make this a subset of functionality under the `update` resolver or make one that's more specific. (`approveSubmission` is a good example of one that's part of updateSubmission. It could technically be a separate resolver, but realistically it's under update because it's an update to the status of submission that has side effects. This feels like a similar case to me)

src/resolvers.ts
```diff
@@ -212,6 +213,66 @@ const resolvers = {
...
+                    },
+                    nameEn: '',
+                    nameJa: '',
+                    mapLatitude: 0,
+                    mapLongitude: 0,
+                    healthcareProfessionalIds: []
+                }             
+            }
+
+            const autofillPlacesInformation
```
ermish, line 245: this looks like logic that should be moved to the service

utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,196 @@
+import puppeteer from 'puppeteer'
+import axios from 'axios'
+import dotenv from 'dotenv'
+
+//Load the api key for google maps
+dotenv.config()
+const apiKey = process.env.GOOGLE_API_KEY
+
+// This uses all the helper functions in order to get the data we want
+export const getFacilityDetailsForSubmission = async (submittedURL: string): Promise<any | null> => {
```
ermish, line 10: This is personal opinion, but I see utils as functionality that's independent of our app logic. (like logging or helper functions) 
Services hold the logic that's specific to our logic. (for example, you wouldn't copy to another project because it doesn't have "facilities") 

I think this could be split into a service and util. The specific code for submissions/domain logic should be in the service, and the generic/reusable google maps logic can be a `googleMapsUtil.ts`

ermish (changes requested):
left some comments that should reshape this PR. I review in more depth afterwards

ermish:
@NabbeunNabi bump in case you missed my review

ermish:
@NabbeunNabi and I talked about this. 
Updated user flow:
- The backend will have a mutation for updating the location and triggering the google search. 
- the front end will have a button on the moderator page to manually trigger the mutation
- the front end can update the button to say something like "rerun" when the data already exists. (it can be rerun, but want to call out that it has already been run)

NabbeunNabi:
@ermish I am going to get to this now. Due to our discussion. With it becoming a mutation where the button will exist in the future. Should I still refactor the locations. Or just fix the mutation name with some comments that help people understand the logic?

ermish:
> @ermish I am going to get to this now. Due to our discussion. With it becoming a mutation where the button will exist in the future. Should I still refactor the locations. Or just fix the mutation name with some comments that help people understand the logic?

if I understand, yes, it should still be moved into the service and an updated mutation

src/resolvers.ts
NabbeunNabi: OK

NabbeunNabi: Makes sense to me

src/resolvers.ts
NabbeunNabi: It is there now!

utils/submissionDataFromGoogleMaps.ts
```diff
@@ -0,0 +1,192 @@
...
+    
+    // regex to extract the full coordinates from url
+    const regexToExtractGoogleMapsCoordinates = /@(-?\d+\.\d+),(-?\d+\.\d+),/
+    const match = redirectedUrl ? redirectedUrl.match(regexToExtractGoogleMapsCoordinates) : null
+
+    if (!match) {
+      return null
+    }
+
+    console.error('No coordinates found in the URL.')
```
ermish, line 31: I think this error is in the wrong spot

ermish (approved):
looks awesome! lots of logic in here. If you remove the console error it's ready to merge

utils/submissionDataFromGoogleMaps.ts
NabbeunNabi: Should be in the if. Silly me

PR #499, opened 2024-07-08
chore: add codeowners file

theyokohamalife:
# What changed

This adds the CODEOWNERS file so GitHub will automatically tag and require a review either from myself or @ermish


ermish (approved):
code ownership!

PR #521, opened 2024-07-16
docs: replace the run API command with the correct one in the README

### LINKED ISSUE #520: DOC: Change wrong command in README
# Summary

Under the title **Run the API** the following command is written `npm run dev:startserver` this will indeed start the server but we want the command that will install the dependencies and generate the graphQL types. 

# Requirements

- [ ] Replace the wrong command with `npm run dev`

# Dependencies

Anissa3005:
Resolves #520 

# What changed
- Under the title **Run the API** I have replaced the `npm run dev:startserver` command with `npm run dev`
- Added extra information how to run the API and what the command does
- Removed unnecessary title from the setting up the database

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,12 +102,6 @@ firebase login
 npm run dev:startlocaldb
 ```
 
-### 3.c Starting the dev server and seeding the emulator
-
-```sh
-npm run dev
-```
-
 This will run until you shut down the instance hitting `ctrl^+C`
 
 ### (Optional) Connecting to Production database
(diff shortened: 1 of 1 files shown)


theyokohamalife:
Looks good! Your PR title is clear, but I think you were trying to follow the syntax for [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary). If so, here is an example of how doc commit (and PR title in the case) messages should be written:

Screenshot 2024-07-16 at 3 08 54 PM

Notice that it's `docs:` and everything is in lowercase except for the file name.

theyokohamalife (approved):

PR #532, opened 2024-07-23
feat: create a mutation for audit logs

### LINKED ISSUE #522: FEAT: Create mutation for audit log
# Summary

Create a mutation for the audit log, so we can add logs to our database.

# Requirements

- [ ] Create a mutation for audit log

# Dependencies

Anissa3005:
Resolves #522 

# What changed

-  Add a mutation for the audit logs, so we are able to add new audit log to our database when certain actions are triggered.
- This function will be called when for example a healthcare professional is created.

# To do

- Add testing, the challenge is where should we tests this when a certain action is triggered, or create a unit test only for this function with a mock database?
- Add validations


src/resolvers.ts
```diff
@@ -250,6 +251,14 @@ const resolvers = {
 
             convertErrorsToGqlErrors(deleteSubmissionResult)
             return deleteSubmissionResult.data
+        },
+
+        createAuditLog: async (_parent: unknown, args: { input: gqlType.CreateAuditLog })
```
ermish, line 256: This actually should not be a mutation exposed to the front end. Instead it will just be logic that is triggered when any service mutation occurs.

src/services/auditLogService.ts
```diff
@@ -0,0 +1,105 @@
+import * as gqlTypes from '../typeDefs/gqlTypes.js'
+// import * as dbSchema from '../typeDefs/dbSchema.js'
+import { ErrorCode, Result } from '../result.js'
+import { dbInstance } from '../firebaseDb.js'
+import { logger } from '../logger.js'
+
+export async function getAuditLogById(id: string): Promise<Result<gqlTypes.AuditLogs>> {
```
ermish, line 7: Realistically I think we will want to query them by latest or by entity type. 
Ex facilities changes in the last month


We might want a get by id later for an undo functionality though but probably not important at this moment

src/resolvers.ts
Anissa3005: I also thought so but my only confusion was that how can I check the data that I put in the database against a schema? So I thought we should call the resolver?

ermish: Do you mean in the test? 
You should be able to use a cypress intercept

ermish: Ohhh I think you mean just saving to the db. We have the dbschema types for that. The whole reason why we have that is because there are some tables and fields in tables that should not be exposed to the public api. Creating Audit logs is a perfect example

Anissa3005: Ah I see, now the dbschema is more clear to me. ☺️I will rework this 🙏

src/services/auditLogService.ts
```diff
@@ -0,0 +1,43 @@
+import * as gqlTypes from '../typeDefs/gqlTypes.js'
+import * as dbSchema from '../typeDefs/dbSchema.js'
+// import { ErrorCode, Result } from '../result.js'
+import { dbInstance } from '../firebaseDb.js'
+import { logger } from '../logger.js'
+
+type CreatedAuditLog = {
+    isSuccesful: boolean
```
Anissa3005, line 8: Should I create this type in the schema so it generates it in gqlType File?

src/services/auditLogService.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export async function createAuditLog(auditLogInput: dbSchema.AuditLog): Promise<CreatedAuditLog> {
+    try {
+        // TO DO: Validations here
+
+        const auditLogRef = dbInstance.collection('auditLogs').doc()
+        const newAuditLogId = auditLogRef.id
+        const newAuditLog = mapGqlEntityToDbEntity(auditLogInput, newAuditLogId)
+
+        await auditLogRef.set(newAuditLog)
```
Anissa3005, line 19: I assume when the set fails it would catch it in the catch block (to much GO the last week)

src/services/auditLogService.ts
ermish: theoretical answer:
your db types should be independent from your api as some fields/tables/entites are not exposed to the client. Audit logs are a good example of something you don't want the user to really know about. To this point, these types should be created either in the service or dbtypes for ones persisted. 

practical answer:
we have an admin panel that will be able to see audit logs, so I think we can add it to the gql schema and use it from there.

src/services/auditLogService.ts
ermish: haha you got it! you can actually create more than one catch for specific errors. So in this example, if you know the error types that might be thrown from the set, you can handle them differently than the generic error handler. 
In this case, I'm not sure how much we would do differently other than something like a retry 🤔

src/typeDefs/dbSchema.ts
```diff
@@ -26,6 +26,12 @@ export type Submission = Omit<gqlType.Submission, 'facility' | 'healthcareProfes
     healthcareProfessionals: gqlType.HealthcareProfessionalSubmission[]
 }
 
+export type AuditLog = Omit<gqlType.AuditLog, 'id' | 'updatedDate'>
+
+export type AuditLogResult = {
+    isSuccesful: boolean
+}
```
Anissa3005, line 33: I made the decision to make the type in the dbSchema since I think we actually never will need to return this to the client. Since the creation of an audit log will happen when another action is triggered like creating healthcare professional

ermish (approved):
audited and looks good!

PR #533, opened 2024-07-24
feat: add names filter to searchProfessionals

### LINKED ISSUE #531: FEAT: Add name filter to searchProfessionals
# Summary
At the moment we don't have a query where we can get a healthcare professional by name. We should add a query in the `searchProfessionals` function.

# Requirements

- [ ] Add query to be able to get a healthcare professional by name

# Dependencies

Anissa3005:
Resolves #531 

# What changed

Added a filter to search professionals by name.

# To Do

-Add testing.

### DIFF
--- a/src/services/healthcareProfessionalService.ts
+++ b/src/services/healthcareProfessionalService.ts
@@ -11,7 +11,7 @@ import { logger } from '../logger.js'
 /**
  * Gets the Healthcare Professional from the database that matches on the id.
  * @param id A string that matches the id of the Firestore Document for the professional.
- * @param firestoreRef An optional reference to the Firestore database within a transaction. If we don't use this during a transaction, we might not get the latest saved data. 
+ * @param firestoreRef An optional reference to the Firestore database within a transaction. If we don't use this during a transaction, we might not get the latest saved data.
  * @returns A Healthcare Professional object.
  */
 export async function getHealthcareProfessionalById(id: string)
@@ -55,7 +55,7 @@ export async function getHealthcareProfessionalById(id: string)
 }
 
 /**
-* This is a search function that will return a list of Healthcare Professionals that match the filters. 
+* This is a search function that will return a list of Healthcare Professionals that match the filters.
 * - At the moment, filters have to match exactly, and there is no fuzzy search.
 * @param filters All the optional filters that can be applied to the search.
 * @returns The matching Healthcare Professionals.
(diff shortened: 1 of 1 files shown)


src/services/healthcareProfessionalService.ts
```diff
@@ -75,10 +75,13 @@ Promise<Result<gqlTypes.HealthcareProfessional[]>> {
         const hasSpecialtiesFilter = filters.specialties && filters.specialties.length > 0
         const hasSpokenLanguagesFilter = filters.spokenLanguages && filters.spokenLanguages.length > 0
         const hasDegreesFilter = filters.degrees && filters.degrees.length > 0
+        const hasNamesFilter = filters.names && filters.names.length > 0
```
ermish, line 78: can remove the `> 0` part

src/services/healthcareProfessionalService.ts
```diff
@@ -134,8 +146,22 @@ Promise<Result<gqlTypes.HealthcareProfessional[]>> {
                 ? professionalsFilteredBySpokenLanguages.filter(professional =>
                     professional.degrees.some(db => filters.degrees?.some(filter => filter == db)))
                 : professionalsFilteredBySpokenLanguages ?? []
+            
+            const professionalsFilteredByName = hasNamesFilter
+                ? professionalsFilteredByDegrees.filter(professionals => {
+                    const filterName = filters.names ? filters.names[0] : false
```
ermish, line 152: should check length otherwise 0 will be undefined when accessing
```suggestion
                    const filterName = filters.names.length ? filters.names[0] : false
```

ermish (changes requested):
almost there!

src/services/healthcareProfessionalService.ts
Anissa3005: Is the length not necessary? Since the filters.names is an array with an object inside? So when I send an empty array it would pass the check but without containing the name?

ermish: I meant `filters.names.length > 0` runs the same as `filters.names.length` in a boolean check

src/services/healthcareProfessionalService.ts
```diff
@@ -145,7 +171,21 @@ Promise<Result<gqlTypes.HealthcareProfessional[]>> {
                     professional.degrees.some(db => filters.degrees?.some(filter => filter == db)))
                 : filteredResults ?? []
 
-            filteredResults = professionalsFilteredByDegrees
+            const professionalsFilteredByName = hasNamesFilter
+                ? professionalsFilteredByDegrees.filter(professionals => {
+                    const filterName = filters.names ? filters.names[0] : false
```
ermish, line 176: should check length here. technically it can exist but be empty and would break if you access by `[0]`

ermish (changes requested):
almost there!

src/services/healthcareProfessionalService.ts
Anissa3005: We mist this one sorry! 🙇‍♀️

src/services/healthcareProfessionalService.ts
Anissa3005: Of course since 0 is falsy 🙇‍♀️

src/services/healthcareProfessionalService.ts
```diff
@@ -93,6 +96,14 @@ Promise<Result<gqlTypes.HealthcareProfessional[]>> {
             searchRef = searchRef.where('degrees', 'array-contains-any', filters.degrees)
         }
 
+        if (shouldFilterByNames) {
+            // Extra check is needed to get the first index of names array filter.
+            // Could not reuse hasNamesFilter due to TypeScript syntax error.
+            if (filters.names && filters.names?.length > 0) {
```
ermish, line 102: Nit: you also don't need the > 0 here
```suggestion
            if (filters.names && filters.names?.length) {
```

ermish (approved):
Ship it!

PR #540, opened 2024-08-06
fix: remove hyphen from docker-compose command

Anissa3005:
# What changed

In previous PR we noticed that the CI was failing due to the old `docker-compose` command.
We updated this to the new `docker compose` command, the CI seems to work again on this branch.

These PR are failing due to the docker command: 
- #533 
- #532


ermish (approved):
ship it! now you're a docker expert 😉

PR #541, opened 2024-08-08
feat: add audit trailing to submissions

### LINKED ISSUE #276: Add audit trails for submissions functions
# Summary

We want to have audit logs of any changes to data. This way we can see when data is add/edited/deleted. 
Ideally, this should also let us undo/rollback changes in the future.

This issue is for the submissions methods from this list

- facilities (#274)
- healthcare (#275) 
- submissions (#276)

# Requirements

- [ ] Add auditing to all the service methods
- [ ] when deleting, add the record that was deleted into the log
    - [ ] ensure it's in a format that can be used to revert a deletion
- [ ] when updating, record the changes in the log
- [ ] when adding, just point to the record id that was created

Depends on #79

Anissa3005:
Resolves #276 
# What changed
- Add audit trailing when submission is created
- Add audit trailing when submission is deleted
- Add audit trailing when submission is updated

# To Do 
-  Create tests for audit trailing

### DIFF
--- a/src/services/submissionService.ts
+++ b/src/services/submissionService.ts
@@ -191,7 +191,7 @@ Promise<Result<gqlTypes.Submission>> => {
             )
 
             if (!createdAuditLog.isSuccesful) {
-                throw new Error(`Faild to create and audit log on ${gqlTypes.ActionType.Create}`) 
+                throw new Error(`Failed to create and audit log on ${gqlTypes.ActionType.Create}`) 
             }
         })
 
--- a/src/services/submissionService.ts
+++ b/src/services/submissionService.ts
@@ -8,13 +8,14 @@ import { createFacility } from './facilityService.js'
 import { createHealthcareProfessional } from './healthcareProfessionalService.js'
 import { validateSubmissionSearchFilters, validateCreateSubmissionInputs } from '../validation/validateSubmissions.js'
 import { logger } from '../../src/logger.js'
+import { createAuditLog } from './auditLogService.js'
 
 /**
  * Gets the Submission from the database that matches the id.
  * @param id A string that matches the id of the Firestore Document for the Submission.
  * @returns A Submission object.
  */
-export const getSubmissionById = async (id: string): Promise<Result<gqlTypes.Submission | undefined>> => {
+export const getSubmissionById = async (id: string): Promise<Result<gqlTypes.Submission>> => {
     try {
         const validationResult = validateIdInput(id)
 
@@ -28,7 +29,7 @@ export const getSubmissionById = async (id: string): Promise<Result<gqlTypes.Sub
 
         if (!dbDocument.exists) {
             return {
-                data: undefined,
+                data: {} as gqlTypes.Submission,
                 hasErrors: true,
                 errors: [{
                     field: 'id',
@@ -51,7 +52,7 @@ export const getSubmissionById = async (id: string): Promise<Result<gqlTypes.Sub
         logger.error(`ERROR: Error retrieving submission by id ${id}: ${error}`)
 
         return {
-            data: undefined,
+            data: {} as gqlTypes.Submission,
             hasErrors: true,
             errors: [{
                 field: 'getSubmissionById',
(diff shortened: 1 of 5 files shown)


src/services/submissionService.ts
```diff
@@ -175,17 +176,38 @@ Promise<Result<gqlTypes.Submission>> => {
...
+            // I should add more specific errors
+            throw new Error('Error adding audit log')
         }
 
+        await batch.commit()
+
+        const createdSubmission = mapDbEntityTogqlEntity(newSubmission)
+
+        // const createdSubmission = await getSubmissionById(newSubmissionId)
+
```
Anissa3005, line 203: I think this is not necessary so for now I commented it out please correct me if I'm wrong

src/services/submissionService.ts
```diff
@@ -175,17 +176,38 @@ Promise<Result<gqlTypes.Submission>> => {
...
-        await submissionRef.set(newSubmission)
+        const batch = dbInstance.batch()
+        
+        batch.set(submissionRef, newSubmission)
+
+        const audit: dbSchema.AuditLog = {
+            actionType: gqlTypes.ActionType.Create,
+            objectType: gqlTypes.ObjectType.Submission,
+            schemaVersion: gqlTypes.SchemaVersion.V1,
+            updatedBy: 'client',
```
Anissa3005, line 187: In this case this is updated by the user, so not sure if I can just hardcode this or not?

src/services/submissionService.ts
```diff
@@ -171,21 +172,42 @@ Promise<Result<gqlTypes.Submission>> => {
...
+            gqlTypes.ObjectType.Submission, 
+            '', 
+            newSubmission
+        )
 
-        // if we didn't get it back or have errors, this is an actual error.
-        if (createdSubmission.hasErrors || !createdSubmission.data) {
-            throw new Error(`${JSON.stringify(createdSubmission.errors)}`)
+        if (!createdAuditLog.isSuccesful) {
+            // I should add more specific errors
```
ermish, line 194: you should add a more specific error 😆

src/services/submissionService.ts
```diff
@@ -171,21 +172,42 @@ Promise<Result<gqlTypes.Submission>> => {
             }
         }
 
+        // => After it's validated we should send already an okay request to the user
```
ermish, line 175: 👍

ermish: should note this is a TODO

src/services/submissionService.ts
```diff
@@ -171,21 +172,42 @@ Promise<Result<gqlTypes.Submission>> => {
...
+
         const submissionRef = dbInstance.collection('submissions').doc()
         const newSubmissionId = submissionRef.id
         const newSubmission = mapGqlEntityToDbEntity(submissionInput, newSubmissionId)
         
-        await submissionRef.set(newSubmission)
+        const batch = dbInstance.batch()
```
ermish, line 181: we should [use a transaction](https://firebase.google.com/docs/firestore/manage-data/transactions#passing_information_out_of_transactions) instead of batch. 
Batch will do it with one request which is nice, but it doesn't handle rolling back an action if any of them fails.

src/services/submissionService.ts
ermish: this is a subjective approach to if you should return the saved object or fetch a new one. 

returning the same has a risk of any side effects that happened aren't picked up, but it saves a whole database query. (for example, maybe there's some other code running in an independent task on a database change)
also, if I remember, we don't get back the id of the created submission without a fetch (but it's been a while so I could be wrong)

src/services/submissionService.ts
ermish: I would just add what action it's failing from when trying to add it

src/services/submissionService.ts
```diff
@@ -450,3 +508,22 @@ const mapDbEntityTogqlEntity = (dbEntity: dbSchema.Submission): gqlTypes.Submiss
 
     return gqlEntity
 }
+
+async function recordAuditLog(
```
ermish, line 512: let's move this into the auditLogService.js

ermish (changes requested):
this is looking pretty clean! a few changes

src/services/submissionService.ts
Anissa3005: I have un commented this.
But we do create id as well, so we have access to it without getting the submission back. 
```
   const newSubmissionId = submissionRef.id
   const newSubmission = mapGqlEntityToDbEntity(submissionInput, newSubmissionId)
```

src/services/submissionService.ts
```diff
@@ -171,11 +172,28 @@ Promise<Result<gqlTypes.Submission>> => {
...
+                gqlTypes.ActionType.Create, 
+                gqlTypes.ObjectType.Submission, 
+                '', 
+                newSubmission,
+                null,
+                t
+            )
+
+            if (!createdAuditLog.isSuccesful) {
+                throw new Error(`Faild to create and audit log on ${gqlTypes.ActionType.Create}`)
```
ermish, line 194: nit spelling
```suggestion
                throw new Error(`Failed to create and audit log on ${gqlTypes.ActionType.Create}`) 
```

ermish (approved):
nice job! looks great!
we definitely will need  to get the tests in there  to validate this functionality

PR #567, opened 2024-08-30
fix: implement isUnderReview boolean change

### LINKED ISSUE #566: FIX: the `approveSubmission` function on the backend doesn't change `underReview` to false
# Summary
This function works beautifully!!! However it should also set `underReview` to `false` if it `isApproved`

# Requirements

- [x] Make this function set `isUnderReview` to false

NabbeunNabi:
Resolves #566 

# What changed
Before it just set a submission to `isApproved` but also should set `isUnderReview` to false if it has been approved.


theyokohamalife (approved):
Makes sense! I would suggest checking that the tests are also testing for this behavior.

PR #571, opened 2024-09-03
fix: reorder logic of creation of submission

### LINKED ISSUE #570: Reorder logic of approving facility then healthcare professional per schema
# Summary
Currently our schema is set up that you need to have a `facilityID` in order to create healthcare professionals. The current logic and function does it in the reverse order.

# Requirements

- [x] The `approveSubmission` function approves a submission, then creates facility, then healthcare professionals with that relation
- [x] Tests are written and passing to make sure this functionality works properly

NabbeunNabi:
Resolves #570 

# What changed
Before the function for `approveSubmission` had the order wrong on what needed to be created when approving a facility. Now a facility is created...and then the healthcare professionals if they need to be. Then finally the submission is changed to `isApproved` This is so that the correct order of logic exists. Also before there was a bug that we never set the data of the request to the `Submission`. Now this is set properly so we can send back the request to the frontend. All the steps should happen in this order or it should not be approved since some some error occurred in the approval process.

# Testing instructions
A test was added to test the functionality of this functionality based on this logic order.

1.  Submission is created
2. Submission is updated to `isUnderReview`
3. Submission is updated with all the necessary values for approval.
4. Submission is approved with no errors.

### DIFF
--- a/__tests__/submissions.test.ts
+++ b/__tests__/submissions.test.ts
@@ -1,12 +1,13 @@
 import { expect, describe, test } from 'vitest'
 import request from 'supertest'
 import { generateRandomCreateSubmissionInput, generateRandomUpdateSubmissionInput } from '../src/fakeData/fakeSubmissions.js'
-import { CreateSubmissionInput, Submission, SubmissionSearchFilters } from '../src/typeDefs/gqlTypes.js'
+import { CreateSubmissionInput, Submission, SubmissionSearchFilters, UpdateSubmissionInput } from '../src/typeDefs/gqlTypes.js'
 import { Error, ErrorCode } from '../src/result.js'
-import { generateSpokenLanguages } from '../src/fakeData/fakeHealthcareProfessionals.js'
+import { generateRandomCreateHealthcareProfessionalInputArray, generateSpokenLanguages } from '../src/fakeData/fakeHealthcareProfessionals.js'
 import { gqlMutation, gqlRequest } from '../utils/gqlTool.js'
 import { gqlApiUrl } from './testSetup.test.js'
 import { logger } from '../src/logger.js'
+import { generateRandomCreateFacilityInput } from '../src/fakeData/fakeFacilities.js'
 
 describe('createSubmission', () => {
     test('creates a new Submission', async () => {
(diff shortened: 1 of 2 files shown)


src/services/submissionService.ts
```diff
@@ -301,35 +301,40 @@ export const approveSubmission = async (submissionId: string): Promise<Result<gq
...
-        currentSubmission.updatedDate = new Date().toISOString()
-
-        await submissionRef.set(currentSubmission, { merge: true })
-
+    
         logger.info(`\nDB-UPDATE: Submission ${submissionId} was approved.`)
-
+        
+        //try creating the facility
+        const createFacilityResult = await createFacility(currentSubmission.facility as dbSchema.Facility)
```
ermish, line 308: This isn't the write input type. It should be the createFacilityInput

src/services/submissionService.ts
```diff
@@ -301,35 +301,40 @@ export const approveSubmission = async (submissionId: string): Promise<Result<gq
...
+        currentSubmission.isUnderReview = false
+        currentSubmission.updatedDate = new Date().toISOString()
 
-        if (createFacilityResult.hasErrors) {
-            approveResult.errors?.concat(createFacilityResult.errors!)
-            return approveResult
-        }
+        await submissionRef.set(currentSubmission, { merge: true })
+
+        //set the data to return to the graphQl response
```
ermish, line 336: nit: this function might not be called from a resolver, so we shouldn't imply it's a graphql response

ermish (changes requested):
just one quick change and make sure it's also working with the latest updates from main around added transactions
I checked through the logic change, and it should be correct to create the facility first👍

__tests__/submissions.test.ts
```diff
@@ -109,6 +110,98 @@ describe('updateSubmission', () => {
...
+
+        // Get the ID of the new Submission
+        const newSubmission = newSubmissionResult.body.data.createSubmission as Submission
+
+        // Mutation to update the Submission
+        const updateSubmissionRequest = {
+            query: updateSubmissionMutation,
+            variables: {
+                id: newSubmission.id,
+                input: generateRandomUpdateSubmissionInput({ isUnderReview: true })
```
theyokohamalife, line 136: Shouldn't `isUnderReview` be `false` if a submission is approved? 🤔

theyokohamalife:
Really great work! It's always good to take the error paths into consideration when writing tests. Some suggestions for additional test cases:
1. When an ID is passed for a submission that doesn't exist
2. When an ID is passed for a submission that is already marked approved
3. When `createFacility` returns errors
4. When `createHealthcareProfessional` returns errors

ermish (approved):
Ship it!

PR #572, opened 2024-09-09
test: add test for audit trailing

### LINKED ISSUE #573: Add tests for audit log function
# Summary

Create integration test for the audit log function.

# Requirements

- [ ] Add happy path test for audit log function

# Dependencies

Anissa3005:
Resolves #573

# What changed
- Add happy path test for audit log function
- Update the docker test command, remove the single quotes https://docs.docker.com/reference/cli/docker/compose/
-  Add a vitest config file to make sure that the testSetup file runs before every test. Without this it caused an error when using the dbInstance

# Testing instructions
- `npm run test:dockerstart`
- `npm run test`


ermish (approved):
It's very weird that we're using vitest on the backend but I think it works 👍 (plus we discuss it before )

PR #588, opened 2024-09-18
chore(lang): add unspecified language choice

### LINKED ISSUE #587: Add unspecified enum type to the locales schema
# Summary
Right now we have submissions that might have locales given, but that is assuming the user submits everything correctly. So with a new submission we should have a unspecified type as well

# Requirements

- [x] Add an unspecified type

NabbeunNabi:
Resolves #587 

# What changed
There was an assumption that the user can accurately report languages. So on a new Healthcare Professional submission it is good to have this type shown as unspecified until the moderator has the chance to confirm.

### DIFF
--- a/src/typeDefs/gqlTypes.ts
+++ b/src/typeDefs/gqlTypes.ts
@@ -236,6 +236,7 @@ export enum Locale {
   ThTh = 'th_TH',
   TlPh = 'tl_PH',
   TrTr = 'tr_TR',
+  Und = 'und',
   ViVn = 'vi_VN',
   ZhCn = 'zh_CN',
   ZhHk = 'zh_HK',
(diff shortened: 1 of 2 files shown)


NabbeunNabi:
@ermish Why did you close this? This was related to adding an enum for an unspecified language type throughout the app that you mentioned in https://github.com/ourjapanlife/findadoc-web/pull/718#discussion_r1763363563

src/typeDefs/gqlTypes.ts
```diff
@@ -236,6 +236,7 @@ export enum Locale {
   ThTh = 'th_TH',
   TlPh = 'tl_PH',
   TrTr = 'tr_TR',
+  Und = 'und',
```
ermish, line 239: nit: could add a comment to explain this is undefined since the shortened version not be intuitive to the reader.

ermish: after talking to @NabbeunNabi it sounds like this is an official recommended value, so I would add that note to the comment

ermish: my bad, this is a generated file and the comment is already on the schema file

ermish (approved):
lets ship it!

PR #642, opened 2024-11-22
fix: refactor healthcare facility relationship to properly update database

### LINKED ISSUE #641: Fix relation functions when updating a facility
# Summary
Currently the relationship when updating a facility are not working as expected. The function to update the healthcare professional was written in a way that we mapped to return an array of the data, but removed the ref. This needs to be fixed. Updating a facility to remove the healthcare professional is written in a way that we make an object that adds the relation to the db due to mapping the fields and updating that. We need to map the fields and then still set the `healthcareProfessionalIds` to the updated ids. Not the relationship.

# Requirements

- [x] Fix the function that updates the HPs by removing the facility they are associated with and only that one as an HP can work at multiple facilties 
- [x] Update the facility to correctly remove the relationship, but have the key value pair be an array of ids, instead of currently putting the array of `Relations` which is what is sent from the frontend due to the schema

NabbeunNabi:
Resolves #641 

# What changed
Before due to some variable names and update order the backend was incorrectly updating the relationships of both the healthcare professional with the facility and the facility relationship to remove the id of the HP. It is now correctly updating both the HP and Facility with the changes.

### DIFF
--- a/src/services/facilityService.ts
+++ b/src/services/facilityService.ts
@@ -220,6 +220,8 @@ export const updateFacility = async (facilityId: string, fieldsToUpdate: Partial
         const dbDocument = await facilityRef.get()
         const dbFacilityToUpdate = dbDocument.data() as dbSchema.Facility
 
+        const originalHealthcareProfessionalIdsForFacility = dbFacilityToUpdate.healthcareProfessionalIds
+
         //let's update the fields that were provided
         MapDefinedFields(fieldsToUpdate, dbFacilityToUpdate)
 
@@ -232,14 +234,13 @@ export const updateFacility = async (facilityId: string, fieldsToUpdate: Partial
                 dbFacilityToUpdate.id,
                 fieldsToUpdate.healthcareProfessionalIds,
                 batch,
-                dbFacilityToUpdate.healthcareProfessionalIds ?? []
+                originalHealthcareProfessionalIdsForFacility ?? []
             )
 
             // if we didn't get it back or have errors, this is an actual error.
             if (healthcareProfessionalUpdateResults.hasErrors || !healthcareProfessionalUpdateResults.data) {
                 throw new Error(`Error updating facility's healthcareProfessionalIds: ${JSON.stringify(healthcareProfessionalUpdateResults.errors)}`)
             }
-
             //let's update the professional with the new facility ids
             dbFacilityToUpdate.healthcareProfessionalIds = healthcareProfessionalUpdateResults.data
         }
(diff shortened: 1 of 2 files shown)


src/services/facilityService.ts
```diff
@@ -374,7 +376,7 @@ export async function updateFacilitiesWithHealthcareProfessionalIdChanges(
             dbFacilityData.updatedDate = new Date().toISOString()
 
             //This will add the record update to the batch, but we don't want to commit at this point. 
-            batch.set(dbFacility.ref, dbFacilityData, { merge: true })
+            batch.set(dbFacilityRef, dbFacilityData, { merge: true })
             logger.info(`\nDB-UPDATE: Updated facility ${dbFacilityData.id} healthcareprofessional relation ids.\n Updated values: ${JSON.stringify(dbFacilityData)}`)
```
ermish, line 380: Why add the variable? This looks like the same functionality and readability

ermish: Ah ok it's for consistency in the other files

src/services/facilityService.ts
```diff
@@ -220,6 +220,8 @@ export const updateFacility = async (facilityId: string, fieldsToUpdate: Partial
         const dbDocument = await facilityRef.get()
         const dbFacilityToUpdate = dbDocument.data() as dbSchema.Facility
 
+        const originalHealthcareProfessionalIdsForFacility = dbFacilityToUpdate.healthcareProfessionalIds
```
ermish, line 223: Before merging, double check that we don't need a deep copy here. If the references are updated the new list be as well

ermish (approved):
Approved! See comments first

src/services/facilityService.ts
NabbeunNabi: The deep copy is happening within the function where this is being used. Would it be safer to be a deep copy in both places. But it is being passed and then deep copied within that function.

NabbeunNabi: I think that it is being handled well on line 294. So I think we are good! 

```
async function processHealthcareProfessionalRelationshipChanges(facilityId: string,
    healthcareProfessionalRelationshipChanges: gqlTypes.Relationship[],
    batch: WriteBatch,
    originalHealthcareProfessionalIds: string[] = [])
    : Promise<Result<string[]>> {
    // deep clone the array so we don't modify the original
    let updatedProfessionalIdsArray = [...originalHealthcareProfessionalIds]

    healthcareProfessionalRelationshipChanges.forEach(change => {
        switch (change.action) {
            case gqlTypes.RelationshipAction.Create:
                updatedProfessionalIdsArray.push(change.otherEntityId)
                break
            case gqlTypes.RelationshipAction.Delete:
                updatedProfessionalIdsArray = updatedProfessionalIdsArray.filter(id => id !== change.otherEntityId)
                break
            default:
                break
        }
    })
    ```

You even made a note of it!

PR #647, opened 2024-12-05
fix: add conditionals to allow updating specific fields

### LINKED ISSUE #644: Fix validations errors when updating a healthcare professional
# Summary

Currently, updating a healthcare professional requires all fields to be provided. Attempting to update only specific fields, such as the name, results in a validation error stating that required fields are missing.

# Requirements

- [ ] Update validation logic to apply checks only to fields included in the update request.

# Dependencies

Anissa3005:
Resolves #644

Previously, updating a healthcare professional required all fields to be provided. If any fields were missing, it caused validation errors.

This PR adds conditionals to the validation logic, allowing updates to one or more specific fields. For example, you can now update only the name field without providing other fields.

# What changed

1. Add conditionals to the `validateUpdateProfessionalInput` function.

# Testing instructions

### DIFF
--- a/src/services/healthcareProfessionalService.ts
+++ b/src/services/healthcareProfessionalService.ts
@@ -691,16 +691,6 @@ function validateUpdateProfessionalInput(input: Partial<gqlTypes.UpdateHealthcar
         return validationResults
     }
 
-    //business rule: at least one facility id is required
-    if (input.facilityIds && input.facilityIds.length < 1) {
-        validationResults.hasErrors = true
-        validationResults.errors?.push({
-            field: 'facilityIds',
-            errorCode: ErrorCode.UPDATEPROFFESIONAL_FACILITYIDS_REQUIRED,
-            httpStatus: 400
-        })
-    }
-
     if (input.names) {
         validateNames(input.names, validationResults)
     }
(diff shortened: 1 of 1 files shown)


src/services/healthcareProfessionalService.ts
```diff
@@ -681,8 +681,18 @@ function validateUpdateProfessionalInput(input: Partial<gqlTypes.UpdateHealthcar
...
+            field: 'input',
+            errorCode: ErrorCode.MISSING_INPUT,
+            httpStatus: 400
+        })
+        return validationResults
+    }
+
     //business rule: at least one facility id is required
-    if (!input.facilityIds || input.facilityIds.length < 1) {
+    if (input.facilityIds && input.facilityIds.length < 1) {
```
ermish, line 695: I don't think this logic really makes sense anymore actually since the facilityIds here are facilityId updates. 
I think the logic here would be to check if the updates are making the facilityId count less than 1.

ermish (changes requested):
nice! I added one comment. Feel free to slack me if it's confusing

ermish (approved):
ship it! :shipit:

PR #660, opened 2024-12-23
fix: update data type of newAuditLog  argument in createAuditLog test

Anissa3005:
# What changed

Previously, there was a bug in the createAuditLog test caused by the newAuditLog argument having the wrong data type. This PR fixes the issue by updating the newAuditLog argument to the correct data type, string.

# Changes Include

1. Updated the newAuditLog argument data type to string in the createAuditLog test.


theyokohamalife (approved):
Thanks for fixing this quickly!

ermish (approved):
LGTM!

PR #670, opened 2024-12-24
feat: add audit trailing to facility service actions create, update and delete

Anissa3005:
Resolves ##274

# Description

This PR adds audit trailing to the facility service actions: create, update, and delete. It also includes some refactoring to improve functionality and address related issues, such as converting batch writes to transactions.

# What changed

1. Added audit trailing to the following actions: 
  -  `createFacility`
  -   `updateFacility`
  -   `deleteFacility`
2. Fixed a missing parameter in the approveSubmission logic when calling the createFacility function.
3. Refactored updateHealthcareProfessionalsWithFacilityIdChanges to use transactions instead of batch writes.

### DIFF
--- a/utils/databaseSeedTool.ts
+++ b/utils/databaseSeedTool.ts
@@ -19,7 +19,7 @@ export const seedDatabase = async () => {
         const facilityIds: string[] = []
 
         for await (const facility of facilities) {
-            const createdFacilityResult = await createFacility(facility)
+            const createdFacilityResult = await createFacility(facility, 'userId')
 
             //we should fail here if we have errors
             if (createdFacilityResult.hasErrors) {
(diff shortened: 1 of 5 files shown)


utils/databaseSeedTool.ts
```diff
@@ -19,7 +19,7 @@ export const seedDatabase = async () => {
         const facilityIds: string[] = []
 
         for await (const facility of facilities) {
-            const createdFacilityResult = await createFacility(facility)
+            const createdFacilityResult = await createFacility(facility, 'userId')
```
ermish, line 22: let's pass in something like `db Seed tool` so we know where it was created

ermish (approved):
awesome! looks good @Anissa3005 ! There's one comment to check first

PR #687, opened 2025-01-14
feat: add resultsCount to searchFacilities

Anissa3005:
Resolves #[Issue number]

# What changed

# Testing instructions


src/typeDefs/schema.graphql
```diff
@@ -13,6 +13,11 @@ type Facility {
   updatedDate: String!
 }
 
+type SearchResultsFacilities {
```
ermish, line 16: should we make this a generic type? like `SearchResults<Facilities>`?

Anissa3005: Ah yes! I just didn't know if this is possible in GraphQl will do some researching. Thank you for the feedback!

src/resolvers.ts
```diff
@@ -17,7 +17,7 @@ const resolvers = {
             return queryResults.data
         },
         facilities: async (_parent: unknown, args: { filters: gqlType.FacilitySearchFilters })
-        : Promise<gqlType.Facility[]> => {
+        : Promise<gqlType.SearchResultsFacilities> => {
```
ermish, line 20: Since we're changing the object returned to the client side, it's going tovbreak the api requests. 
To solve this we should version this endpoint as v2

ermish (changes requested):
Added another comment. Happy to walk through versioning in more detail on slack :)

ermish:
moving to draft so we can versionize it

NabbeunNabi:
This is stale and has not been worked on for 10 months so will close

PR #688, opened 2025-01-15
fix: change facilities relations update on healthcare professional 

### LINKED ISSUE #689: Fix bug where facilities update with the relations instead of the ids
# Summary
Currently the relationship when updating a healthcare facility are not working as expected. The function to update the facility was written in a way that we mapped to return an array of the data, but removed the ref. This needs to be fixed. Updating a healthcare professional to remove the facility relation was written in a way that we make an object that adds the `Relationships` to the db due to mapping the fields and updating that. We need to map the fields and then still set the `facilityIds` to the updated ids. **Not** the `Relationships`.

Similar to #641 the relationships between the healthcare professionals and facilities were not being updated correctly with the correct type in the db, leading to a server error.

The error it was returning was an internal server error due to the database now containing incorrect types and returning them when being queried.

The current behavior is that it is setting 

# Requirements

- [x] Fix the relationship update in the same manner as PR #641

NabbeunNabi:
Resolves #689 

# What changed
Before due to some variable names and update order the backend was
incorrectly updating the relationships of both the healthcare
professional with the facility and the facility relationship to remove
the id of the HP. It is now correctly updating both the HP and Facility
with the changes when updating a Healthcare professional.

Similar fix to the following PR:
- #642
- issue -> #641


theyokohamalife:
@NabbeunNabi You probably don't want to use two different "Resolves" in your PR description, because Github will automatically mark this PR as resolving one that it didn't.

theyokohamalife (approved):
Nice work! This is a good temporary fix, but I think we should consider refactoring `updateHealthcareProfessional` when we have time.

ermish (approved):
LGTM!

PR #690, opened 2025-01-15
BUG: Adding fix for digitalocean build

ermish:
Digital ocean App Platform is not properly picking up our yarn version, so the build is failing. This should be a fix for it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     },
     "homepage": "https://github.com/ourjapanlife/findadoc-server#readme",
     "engines": {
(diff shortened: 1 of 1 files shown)


theyokohamalife:
@ermish Let's remove it from [this line](https://github.com/ourjapanlife/findadoc-server/blob/757a5234f666caf7d9a202fccd2696812cdc4f36/package.json#L100) since we know it's not working.

theyokohamalife (approved):
Approving but please see the small change I suggested.

PR #692, opened 2025-01-18
chore: downgrade yarn to 4.3.1

theyokohamalife:
# What changed

After speaking with customer support at Digital Ocean it was determined that their platform can only support up to Yarn 4.3.1. Unfortunately, this means we have to downgrade in order to get the backend deploying again.


ermish (approved):
LGTM

PR #693, opened 2025-01-19
chore: Migrating yarn to pnp from node_modules

ermish:
Yarn is now using PnP packages instead of node modules and is more stable  now. Let's migrate to use it

This has an additional perk/side effect that it should fix the digitalocean build caching issue.

# What changed
- Migrated to PnP modules. 
- Added a missing insecure reference that pnp already picked up!

# Testing instructions
- run `yarn` and `yarn dev` on your machine. 
- You might need to delete node modules folder


theyokohamalife (approved):
Nice! Hope this fixes things! 🤞🏾

theyokohamalife:
@ermish Please check CI before merging

PR #714, opened 2025-02-20
containerize development server

Anissa3005:
Resolves #[Issue number]

### DIFF
--- a/docker/development/Dockerfile
+++ b/docker/development/Dockerfile
@@ -1,0 +1,25 @@
+FROM node:22.5.1
+
+# Create app directore
+WORKDIR /usr/src/app
+
+# Install app dependencies
+COPY package.json ./
+COPY yarn.lock ./
+COPY .yarnrc.yml ./
+
+# Yarn
+RUN mkdir -p ./.yarn/releases
+COPY .yarn/releases/yarn-*.cjs .yarn/releases/.
+RUN corepack enable
+
+# Install yarn dependencies
+RUN yarn install --immutable --frozen-lockfile
+
+# Copy app source code
+COPY . .
+
+# Run 
+EXPOSE 8080
+
+CMD [ "yarn", "dev" ]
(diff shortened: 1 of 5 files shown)


docker/development/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM node:22.5.1
```
Anissa3005, line 1: At the moment we run this specific node version (same as servers version), would we want to test the app on different versions?

docker/docker-compose-development.yml
```diff
@@ -0,0 +1,27 @@
...
+    environment:
+      - FIRESTORE_EMULATOR_HOST=db:3004
+      - FIRESTORE_PROJECT_ID=find-a-doc-japan
+      - SERVER_PORT=3000
+    ports:
+      - "3000:3000"
+    depends_on:
+      db:
+        condition: service_started
+    volumes:
```
Anissa3005, line 16: The volumes are not working as I want them to be. I would like that we don't have to rebuild the image every time we make a change to the source code or dependencies

firebase.json
```diff
@@ -6,11 +6,13 @@
     "emulators": {
         "firestore": {
             "enabled": true,
-            "port": 3004
+            "port": 3004,
+            "host": "0.0.0.0"
```
Anissa3005, line 10: The host need to be added to be able to run the emulator in the docker container. But is it better to make a env variable of this? instead of hardcoding it?

docker/firebase-emulator/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:alpine
+
```
Anissa3005, line 2: Here I use alpine since we need to download the CLI into the container. This means I need to add Java as well, since this is a requirement to run the emulator. So I thought running alpine was maybe a good choice, so we can run Alpin Linux what is a bit compacter (totally not sure if I'm right here)

PR #737, opened 2025-04-16
feat: Implemented auth 

### LINKED ISSUE #740: feat: Create auth engine
# Summary
Let's set up auth! We want to lock down our endpoints such as our moderator dashboard to be only accessible by certain users.  This is to build out our auth engine. 

# Requirements

Create auth functionality in server
- [ ] Can decrypt auth0 JWT tokens
- [ ] Can pull latest key from JWK server.
- [ ] Enable rotating keys
- [ ] Create an auth function to run in resolvers

### LINKED ISSUE #741: feat: Create auth roles and permissions
# Summary
Let's set up auth! We want to lock down our endpoints such as our moderator dashboard to be only accessible by certain users.  
This is to set up the roles and permissions for our system. 

# Requirements

- [ ] Roles to add: 
    - [ ] admin (for all things, including logs and user management), 
    - [ ] moderator (for the admin dashboard)
    - [ ] user (for general users) 
- [ ] Permissions
    - [ ] read, write, delete for facilities
    - [ ] read, write, delete for healtcare professions
    - [ ] read, write, delete for submissions

# Dependencies

- #470

### LINKED ISSUE #742: feat: add auth to resolvers
# Summary
Let's set up auth! We want to lock down our endpoints such as our moderator dashboard to be only accessible by certain users.  
This is to add auth to each of our endpoints

# Requirements

- [ ] Lock down each endpoint with auth. Even if it's for an anonymous user. 

# Dependencies

#741, #742

ermish:
Resolves #740 
Resolves #741
Resolves #742 

## 🔧 What changed

1. Fixed an issue with the gql schema return types returning nullable objects causing problems on the front end mappings.
2. Updated fastify packages to work properly with auth0
3. Reworked auth library
    - Can decrypt auth0 JWT tokens
    - Can pull latest key from JWK server.
    - Enabled rotating keys
    - Create an auth function to run in resolvers
4. Added auth to every resolver
5. Set up roles and permissions for all use cases for the app

## 🧪 Testing instructions

- validate server runs


src/auth.ts
```diff
@@ -1,61 +1,198 @@
...
+    'delete:users' = 'delete:users'
+}
 
-    await Promise.resolve(() => { 
-        logger.debug('🔐 TODO add Auth0 here')
-    })
+// These are the different permissions or "scopes" that are associated with each role.
+// For example, an admin might have permissions to read, write, and delete healthcare professionals,
+// while a user might only have permission to read their own profile and posts.
+// A user can have separate scopes from their roles, so we need to check both.
```
theyokohamalife, line 41: We should have tests for the admin, moderator, and user roles to validate that they're working as intended, and also prevent unintended changes to permissions in the future.

src/auth.ts
```diff
@@ -1,61 +1,198 @@
...
+            logger.warn(`Auth: ❌ User ${user.sub} is NOT authorized for the required scopes: ${requiredScopes.join(', ')}`)
+            logger.warn(`Auth: User's current roles ${currentUserRoles.join(', ')} and scopes: ${allUserScopes.join(', ')}.`)
+        } else {
+            logger.info(`Auth: ✅ User ${user.sub} is authorized for the required scopes: ${requiredScopes.join(', ')}`)
+        }
 
-    if (!userId) {
+        return hasRequiredScopes
+    } catch (err) {
+        logger.error('Error parsing user object. Have we updated roles in Auth0 that don\'t match our mapping above?:', err)
```
theyokohamalife, line 117: We should probably make this message a bit more explicit as to what mappings we're referring to since we can't see the code from the logger.

src/auth.ts
```diff
@@ -1,61 +1,198 @@
...
+                alg})
+        },
+        // Correcting the 'algorithm' property
+        sign: {
+            algorithm: 'RS256'
+        }
+    })
+
+    //We want the user to be authenticated for every request. This hook will verify the JWT token on every request
+    //This also allows us to retrive the user object from the JWT token
```
theyokohamalife, line 167:
```suggestion
    // This also allows us to retrieve the user object from the JWT token
```

src/auth.ts
```diff
@@ -1,61 +1,198 @@
...
+        request.user = {
+            sub: 'Anonymous User',
+            name: 'Anonymous',
+            email: '',
+            roles: [Role.User],
+            scope: ''
+        } satisfies User
+    })
+
+    logger.debug('🔐 Initialized Auth system')
```
theyokohamalife, line 197:
```suggestion
    logger.debug('🔐 Initialized Auth system.')
```

src/auth.ts
```diff
@@ -1,61 +1,198 @@
...
+            return true
+        }
 
-    const roles: string[] = []
+        if (!user) {
+            logger.warn('Auth: User is not defined')
+            return false
+        }
+        
+        //This will fail if we change auth0 roles and forget to update our mapping.
```
theyokohamalife, line 99:
```suggestion
        // This will fail if we change auth0 roles and forget to update our mapping.
```

src/index.ts
```diff
@@ -1,10 +1,8 @@
 //we want the logger to be started before anything else
```
theyokohamalife, line 1:
```suggestion
// We want the logger to be started before anything else
```

src/server.ts
```diff
@@ -41,7 +41,8 @@ export const createApolloFastifyServer = async (customPort?: number): Promise<st
         healthcheckUrl: '/health'
     })
 
-    //auth integration
+    //Auth integration
```
theyokohamalife, line 44:
```suggestion
    // Auth integration
```

theyokohamalife (commented):
Looks good. Auth is something we need tests for before it gets merged. It might be something @NabbeunNabi could help out with, or I could take a stab at in 2 weeks.

theyokohamalife:
I left a couple nit code suggestions. If you hit the "commit suggestion" button it'll save you from having to rewrite them yourself, but feel free if you'd rather do it that way!

src/auth.ts
NabbeunNabi: I fixed this in my testing branch that I am writing. It has a commit for this specifically if we want to me it @theyokohamalife

src/auth.ts
NabbeunNabi: @theyokohamalife added

NabbeunNabi:
@theyokohamalife all updated and ready

theyokohamalife (approved):
🚢

__tests__/auth.test.ts
```diff
@@ -1,30 +1,288 @@
...
+        const adminUser = {
+            sub: 'adminTest',
+            name: 'Findadoc Admin',
+            email: 'findadocAdmin@admin.com',
+            roles: [Role.Admin],
+            scope: ''
+        }
+
+        // Admins should have access to all scopes
+        expect(authorize(adminUser, [Scope['read:healthcareprofessionals']])).toBe(true)
```
ermish, line 30: since scopes will change, for these tests, let's just test one scope that works, and one that shouldn't in the corresponding tests. 
For admin, `write:logs` is admin only right now, so that's a good one to test it should have it, but moderator's cannot.

__tests__/auth.test.ts
```diff
@@ -1,30 +1,288 @@
...
+import { authorize, buildUserContext, Role, Scope } from '../src/auth'
+import { envVariables } from '../utils/environmentVariables'
+import { FastifyRequest } from 'fastify'
+
+// Mock envVariables to control testing environment
+beforeEach(() => {
+    vi.restoreAllMocks()
+})
+
+describe('authorize()', () => {
```
ermish, line 11: Let's talk on slack, but I think we should reduce these tests down to three tests. And we should also change these to hit the endpoint, not the authorize method. 

1) Hit an endpoint that has auth and successfully validates a user. 
2) Hit an endpoint that has auth and successfully denies an unauthorized user. 
3) Your last test is good to keep to validate that the payload matches our expected format. This is a nice test because that is an external dependency that we can ensure is working + hasn't changed which would break our code. These are usually called dependency tests. 

This will test the endpoint end-to-end works, validate the `authorize` method works, and ensure the end user's experience is as expected.

PR #739, opened 2025-04-18
fix: change urls in README to consistent slug for gitbook

NabbeunNabi:
Resolves #738

- [x] PR title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) specs

## 🔧 What changed
The links in the `README` were updated to the correct urls for the second and third link

This is previous. If you loo at the changes I updated the links to have the correct slugs for the `Quickstart the API` and `Run the API`
[📒 Welcome to Find a Doc, Japan!](https://documentation.findadoc.jp/development)
[🚅 Quickstart the API](https://app.gitbook.com/o/Hmir5Cugknp7uJaXBpz1/s/8SwzLo8dTmkYT8nK5WXz/start-coding/quickstart-the-frontend-and-backend)
[🚀 Run the API](https://app.gitbook.com/o/Hmir5Cugknp7uJaXBpz1/s/8SwzLo8dTmkYT8nK5WXz/start-coding/quickstart-the-frontend-and-backend/running-the-api-locally)

## 🧪 Testing instructions
Click the links in the read me and it shouldn't ask you to log in

PR #757, opened 2025-04-30
fix: change localhost to use ip address

NabbeunNabi:
This change bypasses the lookup that can cause devs to have issues with the loopback network interface

Resolves #739

## 🔧 What changed
Pointing directly to the IP address instead of using the loopback network interface

## 🧪 Testing instructions
Run the backend locally and see it still works


ermish:
There's a chance this might mess up auth URLs as auth0is pointed to localhost instead of an IP. Since it will only affect locally, I'm good to merge it and hotfix if there are any gotchas 👍

ermish (approved):

PR #780, opened 2025-05-13
feat: implement updated CORS for frontend deploy previews

### LINKED ISSUE #779: Update cors for deploy previews on netlify
# Summary
On the frontend we want to be able to see deploy previews. This is not possible currently in the CORS setup because we have with credentials to true. This disallows wildcard domains and they do not pass the preflight check. So we need to introduce a callback in order to match with regex if it is the deploy preview site.

# Requirements

- [x] Update CORS to allow deploy preview with a match and callback

NabbeunNabi:
Resolves #779 

## 🔧 What changed
Before our cors had a wildcard domain for . This does not work with fastify and `CORS` because the `*` wildcard only works when credentials is false.

This uses a function that takes a callback that returns true or false whether the origin is allowed by checking an array without the flag and a match with regex to match our deployment preview URLS

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -21,7 +21,19 @@ export const createApolloFastifyServer = async (customPort?: number): Promise<st
 
     //cors is a middleware that allows us to make requests from the a different url (findadoc.jp) to our server (api.findadoc.jp)
     await fastify.register(corsPlugin, {
-        origin: [envVariables.websiteURL(), 'localhost', 'http://localhost:3000', '*findadoc.netlify.app', 'https://www.findadoc.jp'],
+        origin: (origin: string | undefined, cb: (err: Error | null, allow: boolean) => void): void => {
+            const allowedOrigins = [
+                envVariables.websiteURL(),
+                'http://localhost:3000',
+                'https://www.findadoc.jp'
+            ]
+
+            const isAllowed =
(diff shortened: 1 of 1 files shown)


src/server.ts
```diff
@@ -21,7 +21,25 @@ export const createApolloFastifyServer = async (customPort?: number): Promise<st
...
     await fastify.register(corsPlugin, {
-        origin: [envVariables.websiteURL(), 'localhost', 'http://localhost:3000', '*findadoc.netlify.app', 'https://www.findadoc.jp'],
+        origin: (origin: string | undefined, cb: (err: Error | null, allow: boolean) => void): void => {
+            const allowedOrigins = [
+                envVariables.websiteURL(),
+                'http://localhost:3000',
+                'https://www.findadoc.jp'
+            ]
+
+            // Always allow requests with no origin (such as server to itself)
```
ermish, line 31: Do we actually need this?

ermish (approved):
Approved but I'm wondering if we need the ability to call itself

src/server.ts
NabbeunNabi: No. Do not need it. Thought it would be there just as like a safeguard in the future. But probably less origins is better just in case until necessary. So I have removed it.

PR #800, opened 2025-06-10
refactor: add autofillPlaceFromSubmissionUrl as part of submission type

### LINKED ISSUE #789: Add `autofillPlaceFromSubmissionUrl` as part of submission type
# Summary
We now have the ability to autofill some facility details from a frontend call. But we want to also have this be part of the submission and automatically set to false. Then when true it will update it and only update it twice. When this happens it will also allow the frontend to disable the button to not even call the api if this field is already true

# Requirements

- [ ] Update `Submission` type to include `autofillPlaceFromSubmissionUrl`
- [ ] Add logic that it only calls the function to autofill if it isn't already true
- [ ] Make sure that the backend sends back this data as part of the payload

matthewkurtti:
Resolves #789 

## 🔧 What changed

In the graphql schema (`gqlTypes.ts` and `schema.graphql`), added an `autofillPlaceFromSubmissionUrl `boolean field to the Submission type in order to keep track of what submissions have had the google places autofill call. When submissions are created, the field has a default value of false. In `submissionService.ts`, if the autofill call is made and successful, before the results are returned, the submissions `autofillPlaceFromSubmissionUrl` is switched to true

## 🧪 Testing instructions

First, from apollo server, when you start the firebase dev server, verify that all the submission's `autofillPlaceFromSubmissionUrl ` field are false by default. Then, try calling an updateSubmission Mutation and see what is returned


src/services/submissionService.ts
```diff
@@ -604,6 +608,7 @@ export async function deleteSubmission(id: string, updatedBy: string)
 export function mapGqlEntityToDbEntity(input: gqlTypes.CreateSubmissionInput, newId: string): dbSchema.Submission {
     return {
         id: newId,
+        autofillPlaceFromSubmissionUrl: false,
```
NabbeunNabi, line 611: This should follow the same pattern as those using input

src/typeDefs/schema.graphql
```diff
@@ -96,6 +96,7 @@ input HealthcareProfessionalSearchFilters {
 
 type Submission {
   id: ID!
+  autofillPlaceFromSubmissionUrl: Boolean!
```
NabbeunNabi, line 99: I would say it will be the same here

src/services/submissionService.ts
matthewkurtti: @NabbeunNabi if it follows the input pattern, does that mean I should add `autofillPlaceFromSubmissionUrl` to the `CreateSubmissionInput` type as well?

matthewkurtti: Done! (I think)

matthewkurtti:
Made the modifications to address your requested changes Will. But now, the firebase database isn't populating any values for some reason. I'll need to mess with it some more

NabbeunNabi:
> Made the modifications to address your requested changes Will. But now, the firebase database isn't populating any values for some reason. I'll need to mess with it some more 

Might have to look in the Faker data

NabbeunNabi (approved):
Great work. This looks good. Onto @ermish

PR #817, opened 2025-06-19
refactor: add additionalInfoForPatients field to hp

### LINKED ISSUE #802: Add an `additionalInfoForPatients` field to healthcare professionals
# Summary
We have some clinics that want us to display additional information for patients and we need to be able to collect this information to show to users of our app. An example would be that they want us to display that they are only open to English patients from 12:00-14:00 on Tuesdays

# Requirements

- [ ] Add this field to healthcare professionals where it would need be used in a submission and in an existing healthcare professional
- [ ] The type is string and is required but an empty string is acceptable
- [ ] Researched how to run a `cron` job to update our current database to add these fields to our current documents that do not already have this key
- [ ] Add a cron job to update our records so going forward we can add this context in submissions in our db and existing healthcare
- [ ] Discuss if versioning is necessary. Currently as of 6/15/2025 @NabbeunNabi does not think so since it is a new field that is not being called by the frontend so will not break anything

# Implementation details
- The following permalinks will give you an idea of where it needs to be updated and added
https://github.com/ourjapanlife/findadoc-server/blob/130a07486f814be9747df0a2ff02b999f56c8e98/src/typeDefs/schema.graphql#L65
https://github.com/ourjapanlife/findadoc-server/blob/130a07486f814be9747df0a2ff02b999f56c8e98/src/typeDefs/schema.graphql#L74
https://github.com/ourjapanlife/findadoc-server/blob/130a07486f814be9747df0a2ff02b999f56c8e98/src/typeDefs/schema.graphql#L53

matthewkurtti:
Resolves #802

## 🔧 What changed

Added an `additionalInfoForPatients` field to hp schema so that we can display additional information that clinics want patients to know about. It's an optional field that by default will be an empty string

## 🧪 Testing instructions

Try starting the local db and local server and see if you can see the new field in the fake hp data (it should be there as empty strings). You can also test the update by trying to send an update request to make a new string for the additionalInfoForPatients field

### DIFF
--- a/__tests__/healthcareProfessional.test.ts
+++ b/__tests__/healthcareProfessional.test.ts
@@ -271,6 +271,7 @@ export const createHealthcareProfessionalMutation = `mutation test_createHealthc
         acceptedInsurance
         createdDate
         updatedDate
+        additionalInfoForPatients
     }
 }`
 
--- a/__tests__/healthcareProfessionalValidations.test.ts
+++ b/__tests__/healthcareProfessionalValidations.test.ts
@@ -23,7 +23,8 @@ describe('error path: validations healthcare professionals', () => {
                         degrees: testData.generateDegrees(),
                         specialties: testData.generateSpecialties(),
                         spokenLanguages: testData.generateSpokenLanguages(),
-                        acceptedInsurance: [testData.generateAcceptedInsurance()]
+                        acceptedInsurance: [testData.generateAcceptedInsurance()],
+                        additionalInfoForPatients: ''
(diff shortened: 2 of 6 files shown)


matthewkurtti:
Just FYI, I still need to do some research into how to run a cron job to update our current database to add these fields to our current documents that do not already have this key

NabbeunNabi (approved):
Nice! Onto @ermish

ermish (approved):
Ship it

PR #818, opened 2025-06-20
refactor: change international to travel insurance

NabbeunNabi:
Update insurance type to travel from international since people do not know international and they are the same


ermish (changes requested):
these are not the same! Let's add both!

ermish (approved):
pretty massive, risky change, but I think we should be safe. :shipit:

Issue #830, opened 2025-07-03
limit firebase query to <30


Ghezz9:
# Summary

@ermish @NabbeunNabi 

Starting from this task

https://github.com/orgs/ourjapanlife/projects/17/views/1?filterQuery=query&pane=issue&itemId=107058413&issue=ourjapanlife%7Cfindadoc-web%7C1138

We decided to avoid the issues about query limit is 30 in a firebase and adding a pagination in backend.
Firstly

**Why we want to avoid the query limit issue (max 30 items in Firestore in queries) and add backend pagination?**
Firestore imposes a strict limit on the number of items that can be passed `in` a single in or `array-contains-any` query filter, currently capped at 30. If we exceed this limit, queries fail and users experience broken functionality.
By avoiding large queries that exceed this limit, we:

- Prevent runtime errors and failed data loads
- Improve query performance by reducing large, expensive queries
- Enable a more scalable and maintainable approach through pagination

Adding backend pagination ensures:

- The server returns data in manageable chunks, reducing load and response time
- The frontend can request only the data it needs per page, improving UX and performance
- We can handle large datasets gracefully without overwhelming Firestore or the client

Overall, this approach improves system robustness, scalability, and user experience by respecting Firestore limits and implementing best practices for data fetching.

I think the two query that are giving to us this problem

`healthcareProfessionalService.ts
const professionalsQuery = dbInstance.collection('healthcareProfessionals').where('id', 'in', professionalRelationshipsToUpdate.map(f => f.otherEntityId))`

`facilityService.ts
const facilityQuery = dbInstance.collection('facilities').where('id', 'in', facilitiesToUpdate.map(f => f.otherEntityId))`

I'm saying this because i've found on Firebase documentation where could be the limit of 30

https://firebase.google.com/docs/firestore/query-data/queries#in_and_array-contains-any

I will explain better in the next hours but after one day and after talked with Phil one possible solution is:
**Batch the queries on the backend**
Instead of querying all IDs at once, split the array of IDs into chunks of max 30 items.
For each batch:
Run the query with `.where('id', 'in', batchOf30Ids)` then collect the partial results then merge all results and return to the frontend.
**Add backend pagination**
Support paginating the merged results or applying pagination before querying batches, so the frontend receives data in manageable pages i think this will reduces the amount of data transferred and will improves performance.

At the end i would say that the best approach could be:
**Why batching + pagination?**
- Batching respects Firestore’s 30-item in filter limit.
[...]

PR #831, opened 2025-07-03
feat: batching queries and create pagination

Ghezz9:
Resolves [#830]

## 🔧 What changed

Still in working, i'm using batching and for now i've changed only `const professionalsQuery = dbInstance.collection('healthcareProfessionals').where('id', 'in', professionalRelationshipsToUpdate.map(f => f.otherEntityId))`.
same as for facilitiesQuery.

Before pushing the PR i've watched again @ermish screen ( here https://github.com/ourjapanlife/findadoc-web/issues/1138) and i've saw the problem wasn't on those line but the error was:
`ARRAY-CONTAINS-ANY support up 30 comparision`, i've checked again `FacilityService.ts` and `healthcareProfessionalService.ts` and i've understood that i wasn't fixing the right line but i have to change this

image

Those line has `array-contains-any` so it is 100% where is the problem, i would ask in the first commit where you advice to me to put the function `chunkArray` because now i need to batch also those lines.

UPDATE from 4th commit, still don't know how to test with hundreds HC/Facilities but if i go to moderation page and click on a submission and try to select exist Facility what i can see from terminal is this

image

i can see `healthcareProfessionalIds` that is were i used batch, app still works

## 🧪 Testing instructions

I would ask also how can i tested this in a good way? i've seen some scripts on `package.json` but i'm not sure about which is best way.
Obviously after first commit app still work perfectly.

### DIFF
--- a/src/services/healthcareProfessionalService.ts
+++ b/src/services/healthcareProfessionalService.ts
@@ -584,10 +584,27 @@ export async function updateHealthcareProfessionalsWithFacilityIdChanges(
             }
         }
 
-        const professionalsQuery = dbInstance.collection('healthcareProfessionals').where('id', 'in', professionalRelationshipsToUpdate.map(f => f.otherEntityId))
+        const MAX_BATCH_SIZE = 30
+
+        function chunkArray<T>(arr: T[], size: number): T[][] {
+            return Array.from({ length: Math.ceil(arr.length / size) }, (_, i) =>
+                arr.slice(i * size, i * size + size))
+        }
+
+        const allProfessionalIds = professionalRelationshipsToUpdate.map(f => f.otherEntityId)
+        const chunks = chunkArray(allProfessionalIds, MAX_BATCH_SIZE)
+
+        // A Firestore transaction requires all reads before any writes — esegui tutte le query prima
+        const querySnapshots = await Promise.all(
+            chunks.map(chunk =>
+                dbInstance.collection('healthcareProfessionals').where('id', 'in', chunk).get())
+        )
+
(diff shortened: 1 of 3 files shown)


src/services/healthcareProfessionalService.ts
```diff
@@ -584,10 +585,22 @@ export async function updateHealthcareProfessionalsWithFacilityIdChanges(
...
+                dbInstance.collection('healthcareProfessionals').where('id', 'in', chunk).get())
+        )
+
+        const allProfessionalDocuments = querySnapshots.flatMap(snapshot => snapshot.docs)
+        //const professionalsQuery = dbInstance.collection('healthcareProfessionals').where('id', 'in', professionalRelationshipsToUpdate.map(f => f.otherEntityId))
         // A Firestore transaction requires all reads to happen before any writes, so we'll query all the professionals first.
-        const allProfessionalDocuments = await professionalsQuery.get()
-        const dbProfessionalsToUpdate = allProfessionalDocuments.docs.map(d => ({
+        //const allProfessionalDocuments = await professionalsQuery.get()
+        const dbProfessionalsToUpdate = allProfessionalDocuments.map(d => ({
```
NabbeunNabi, line 603: can you name this `document` instead of `d`

NabbeunNabi (commented):
@Ghezz9 looks good to me. What do you think @ermish

I think since this is part of the `e2e` flow we should have some tests to make sure that this works properly. I care less about unit tests than the whole flow. I think Phil agrees

Ghezz9:
I Think we can close this PR because i've already implemented batch query in this one

#834

PR #834, opened 2025-07-06
feat: pagination for backend

### LINKED ISSUE #830: limit firebase query to <30

Ghezz9:
Resolves #830

## 🔧 What changed

All my changed is in this Google docs

https://docs.google.com/document/d/14r7OL8s_WD8jlNLEy7aRx5oEwDltXlHUdMcqou8Zr3k/edit?usp=sharing
## 🧪 Testing instructions

Change function like `export function generateRandomCreateHealthcareProfessionalInputArray()` and add more than 30 HealthcareProfessionals


src/services/facilityService.ts
```diff
@@ -62,64 +63,223 @@ export const getFacilityById = async (id: string)
...
-            searchRef = searchRef.where('nameJa', '==', filters.nameJa)
-        }
+            type ComparablePrimitive = string | number | boolean
+
+            // Order the results in memoryy
+            if (filters.orderBy && Array.isArray(filters.orderBy)) {
+                // Helper function to compare two defined values.
+                const comparePrimitiveValues = 
+                (valA: ComparablePrimitive, valB: ComparablePrimitive): number => {
+                    if (valA < valB) {
```
ermish, line 133: can you make this more readable? what is 1,-1,0 mean for example

src/resolvers.ts
```diff
@@ -119,9 +164,31 @@ const resolvers = {
             const matchingSubmissionsResult = await submissionService.searchSubmissions(args.filters)
 
             convertErrorsToGqlErrors(matchingSubmissionsResult)
+
             return matchingSubmissionsResult.data
+        },
+
+        submissionsTotalCount: async (_parent: unknown, args: { filters: gqlType.SubmissionSearchFilters },
```
ermish, line 171: same here

PR #841, opened 2025-07-08
Feat: add 'payment' field to facility type

### LINKED ISSUE #832: Feat: add 'payment' field to facility type
# Summary

Now that we are adding more healthcare professionals and facility to findadoc, its been discovered that many facilities vary on the types of payment options that are available. This seems like useful information for patients to know during their search, so let's try to add this field to the facility type

# Requirements

- [ ] Make a new type called 'payment' that includes common payments (cash, visa, master, etc)
- [ ] Make a new field in facility that is of type 'payment'
- [ ] Confirm that the schema is producing this in the firebase db with screenshots

# Dependencies

matthewkurtti:
Resolves #832

## 🔧 What changed

This PR is a work in progress. It will add a payment field to facility type with the goal of letting users know what payment options there are at facilities

So far, only have created credit card option enum

## 🧪 Testing instructions

Coming soon

### DIFF
--- a/src/typeDefs/gqlTypes.ts
+++ b/src/typeDefs/gqlTypes.ts
@@ -78,15 +78,6 @@ export type CreateSubmissionInput = {
   spokenLanguages?: InputMaybe<Array<Locale>>;
 };
 
-export enum CreditCardOption {
-    Visa,
-    Master,
-    Amex,
-    Jcb,
-    Diners,
-    AmericanExpress
-}
-
 export enum Degree {
   Cnm = 'CNM',
   Dc = 'DC',
@@ -124,6 +115,7 @@ export type Facility = {
   mapLongitude: Scalars['Float']['output'];
   nameEn: Scalars['String']['output'];
   nameJa: Scalars['String']['output'];
+  payment: Array<Record<PaymentType, PaymentTypeInfo>>;
   updatedDate: Scalars['String']['output'];
 };
 
@@ -358,6 +350,35 @@ export type OrderBy = {
 export enum OrderDirection {
   Asc = 'asc',
   Desc = 'desc'
+}
+
+export enum PaymentBrand {
+    CASH,
+    VISA,
+    MASTER,
+    AMEX,
+    DinersClub,
+    JCB,
+    PayPay,
+    nanaco,
+    JDEBIT,
+    Suica
+}
+
+export type PaymentBrandInfo = {
+    label: Scalars['String']['output'];
+    icon: Scalars['String']['output'];
+}
+
+export enum PaymentType {
+    CASH,
+    CreditCard,
+    Electronic
+}
+
+export type PaymentTypeInfo = {
+    label: Scalars['String']['output'];
+    brands: Array<Record<PaymentBrand, PaymentBrandInfo>>
 }
 
 export type PhysicalAddress = {


src/typeDefs/gqlTypes.ts
```diff
@@ -78,6 +78,15 @@ export type CreateSubmissionInput = {
...
 };
 
+export enum CreditCardOption {
+    Visa,
+    Master,
+    Amex,
+    Jcb,
+    Diners,
+    AmericanExpress
+}
```
NabbeunNabi, line 88: This should be added to the schema so it generates similar to how the degrees do below it.

We autogenerate our types based on `graphql.schema` using codegen which you can see see we run with `yarn generate`

NabbeunNabi (changes requested):
Good start @matthewkurtti. Left a comment for you

matthewkurtti:
@NabbeunNabi  could you take a look at my most recent commit? I added the changes for payment that you and @ermish talked about during our last stand up. I took your example and tried to make the types and enums based off that. If it looks good, I'll start using this template and adding to the graphql.schema and other files

src/typeDefs/gqlTypes.ts
```diff
@@ -351,6 +352,35 @@ export enum OrderDirection {
...
+
+export enum PaymentType {
+    CASH,
+    CreditCard,
+    Electronic
+}
+
+export type PaymentTypeInfo = {
+    label: Scalars['String']['output'];
```
NabbeunNabi, line 380: I would change this to `paymentType`. But with codegen and `graphql.schema` it does not generate records. We have to manually write the records ourselves and probably put it on the backend and frontend

src/typeDefs/gqlTypes.ts
```diff
@@ -351,6 +352,35 @@ export enum OrderDirection {
...
+export type PaymentBrandInfo = {
+    label: Scalars['String']['output'];
+    icon: Scalars['String']['output'];
+}
+
+export enum PaymentType {
+    CASH,
+    CreditCard,
+    Electronic
+}
```
NabbeunNabi, line 377: For this one add any others you can think of as well. And maybe me more clear about what Electronic means. Cuz to me. Suica is not really electronic and not really a credit card

src/typeDefs/gqlTypes.ts
```diff
@@ -351,6 +352,35 @@ export enum OrderDirection {
...
+    PayPay,
+    nanaco,
+    JDEBIT,
+    Suica
+}
+
+export type PaymentBrandInfo = {
+    label: Scalars['String']['output'];
+    icon: Scalars['String']['output'];
+}
```
NabbeunNabi, line 371: I think for this we can add this in the record later for the frontend when we want to map and display logos. Does not need to be in the schema

src/typeDefs/gqlTypes.ts
```diff
@@ -351,6 +352,35 @@ export enum OrderDirection {
...
+    VISA,
+    MASTER,
+    AMEX,
+    DinersClub,
+    JCB,
+    PayPay,
+    nanaco,
+    JDEBIT,
+    Suica
+}
```
NabbeunNabi, line 366: There has to be more than this. Maybe the excel sheet I shared doesn't have more listed. But if we expand services we could use more brands. Maybe not go overboard. But Pasmo and Suica are usually interchangeable. Things like that

NabbeunNabi (changes requested):

PR #847, opened 2025-07-23
fix: update mapping to set additional info to empty string instead of undefined

NabbeunNabi:
Resolves #[Issue number]

## 🔧 What changed
Before we casted the type as a string. This would cause it to try to add undefined as the value to the key in firestore which is not allowed. So now we set it to an empty string if undefined. This logic works for firestore and our frontend as we only display notes if it is truthy


ermish (approved):
One string fix

PR #851, opened 2025-07-28
fix: set default for autofillGoogleMapsURL to false as when a user submits it will be false

NabbeunNabi:
Resolves #[Issue number]

### DIFF
--- a/src/services/submissionService.ts
+++ b/src/services/submissionService.ts
@@ -628,7 +628,7 @@ export async function deleteSubmission(id: string, updatedBy: string)
 export function mapGqlEntityToDbEntity(input: gqlTypes.CreateSubmissionInput, newId: string): dbSchema.Submission {
     return {
         id: newId,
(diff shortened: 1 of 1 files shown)


src/services/submissionService.ts
```diff
@@ -628,7 +628,7 @@ export async function deleteSubmission(id: string, updatedBy: string)
 export function mapGqlEntityToDbEntity(input: gqlTypes.CreateSubmissionInput, newId: string): dbSchema.Submission {
     return {
         id: newId,
-        autofillPlaceFromSubmissionUrl: input.autofillPlaceFromSubmissionUrl as boolean,
+        autofillPlaceFromSubmissionUrl: false,
```
ermish, line 631: shouldn't this be 
```suggestion
        autofillPlaceFromSubmissionUrl: input.autofillPlaceFromSubmissionUrl ?? false,
```

ermish (approved):
default approval false. updated to true

PR #856, opened 2025-07-29
fix: change permission scope for creating submission

NabbeunNabi:

## 🔧 What changed
The auth in our backend did not match our store which disallowed generic users to create submissions in prod

### DIFF
--- a/__tests__/auth.test.ts
+++ b/__tests__/auth.test.ts
@@ -99,7 +99,7 @@ describe('authorize()', () => {
         // users should have access to user scopes
         expect(authorize(adminUser, [Scope['read:healthcareprofessionals']])).toBe(true)
         expect(authorize(adminUser, [Scope['read:facilities']])).toBe(true)
-        expect(authorize(adminUser, [Scope['write:submissions']])).toBe(true)
+        expect(authorize(adminUser, [Scope['create:submissions']])).toBe(true)
         expect(authorize(adminUser, [Scope['read:profile']])).toBe(true)
     })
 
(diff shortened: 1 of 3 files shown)


src/auth.ts
```diff
@@ -45,6 +46,7 @@ const roleScopes: Record<Role, Scope[]> = {
         Scope['read:healthcareprofessionals'], Scope['write:healthcareprofessionals'], Scope['delete:healthcareprofessionals'],
         Scope['read:facilities'], Scope['write:facilities'], Scope['delete:facilities'],
         Scope['read:submissions'], Scope['write:submissions'], Scope['delete:submissions'],
+        Scope['create:submissions'],
```
ermish, line 49: no need to add create permission to any of the roles as `write` should already include this. `Create` is `write` without `update`.

__tests__/auth.test.ts
```diff
@@ -213,6 +214,7 @@ describe('roleScopes invariant: no accidental changes', () => {
             Scope['read:healthcareprofessionals'], Scope['write:healthcareprofessionals'], Scope['delete:healthcareprofessionals'],
             Scope['read:facilities'], Scope['write:facilities'], Scope['delete:facilities'],
             Scope['read:submissions'], Scope['write:submissions'], Scope['delete:submissions'],
+            Scope['create:submissions'],
```
ermish, line 217: let's remove this out of the non-user roles

src/resolvers.ts
```diff
@@ -251,7 +251,7 @@ const resolvers = {
         createSubmission: async (_parent: unknown, args: {
             input: gqlType.CreateSubmissionInput
         }, context: UserContext): Promise<gqlType.Submission> => {
-            const isAuthorized = authorize(context.user, [Scope['write:submissions']])
+            const isAuthorized = authorize(context.user, [Scope['write:submissions']]) || authorize(context.user, [Scope['create:submissions']])
```
ermish, line 254: double check, but I'm pretty sure authorize can take an array or Scopes instead of having to call it multiple times

ermish (approved):
permission scope  = authorized

PR #869, opened 2025-08-08
refactor: refactor BE pagination after EVENT fix

Ghezz9:
Resolves
No issues relate but this frontend PR https://github.com/ourjapanlife/findadoc-web/pull/1439

## 🔧 What changed

- Changed the schema adding the 2 new Type
- Explain when and what is the difference beetween `baseQueryResult.list` and `baseQueryResult.query` 
- moved helper function, only moved helper function related to `Submission, Facility`, the one on `HealthcareProfessional` is little different and i didn't know if it was a good idea to move because it's related only to `HealthcareProfessional`
- Explain the logic of `Comparison` method
- Still not change `vitest.config.ts` i prefer do in a different PR

### DIFF
--- a/utils/searchSubmissionsQueryUtils.ts
+++ b/utils/searchSubmissionsQueryUtils.ts
@@ -64,7 +64,7 @@ Promise<{ query?: Query<DocumentData>, list?: gqlTypes.Submission[], hasErrors: 
                 allGqlSubmissions = allGqlSubmissions.filter(s => s.updatedDate === filters.updatedDate)
             }
 
-            const orderCriteria = (filters.orderBy && Array.isArray(filters.orderBy) && filters.orderBy.length > 0)
+            const orderCriteria = (filters.orderBy && Array.isArray(filters.orderBy) && filters.orderBy.length)
                 ? filters.orderBy
                 : [{ fieldToOrder: 'createdDate', orderDirection: gqlTypes.OrderDirection.Desc }]
             


utils/searchFacilityQueryUtils.ts
```diff
@@ -63,70 +63,11 @@ Promise<{ query?: Query<DocumentData>, list?: gqlTypes.Facility[], hasErrors: bo
...
-                })
-            } else {
-                // Default in-memory sorting: sort by 'createdDate' in descending order.
-                allGqlFacilities.sort((facilityA, facilityB) => {
-                    const createdDateA = new Date(facilityA.createdDate)
-                    const createdDateB = new Date(facilityB.createdDate)
-                    return createdDateB.getTime() - createdDateA.getTime()
-                })
-            }
+            const orderCriteria = (filters.orderBy && Array.isArray(filters.orderBy) && filters.orderBy.length > 0)
```
ermish, line 66: can simplify length check. Also, not sure we need to check if it's an array
```suggestion
            const orderCriteria = (filters.orderBy && Array.isArray(filters.orderBy) && filters.orderBy.length)
```

utils/searchFacilityQueryUtils.ts
```diff
@@ -63,70 +63,11 @@ Promise<{ query?: Query<DocumentData>, list?: gqlTypes.Facility[], hasErrors: bo
...
+            ? filters.orderBy
+            : [{ fieldToOrder: 'createdDate', orderDirection: gqlTypes.OrderDirection.Desc }]
+        
+            allGqlFacilities = sortArrayByOrderCriteria(allGqlFacilities, orderCriteria as any)
```
ermish, line 70: I think this should have a concrete type, not `any`

utils/arrayUtils.ts
```diff
@@ -11,4 +29,61 @@ export function removeDuplicates<T>(inputArray: T[]): T[] {
...
+export function sortArrayByOrderCriteria<T extends Record<string, any>>(
+    arrayToSort: T[],
+    orderBy: Array<{ fieldToOrder: keyof T, orderDirection: 'ASC' | 'DESC' }> | null | undefined
+): T[] {
+    // If no sorting criteria are provided, return the original array without sorting.
+    if (!orderBy || !Array.isArray(orderBy) || orderBy.length === 0) {
+        return arrayToSort
+    }
+
+    return arrayToSort.sort((a, b) => {
```
ermish, line 53: much more readable 👍

ermish (changes requested):
looks great! just some small nits. 
We also need to make sure this doesn't break the front end when it's merged. Can you merge and check it?

ermish:
I think this is in draft, so updating it

PR #876, opened 2025-08-15
fix: fix dates for healthcareProfessional

Ghezz9:
Resolves #[No issue number]

## 🔧 What changed

Fixed `Date` related to healthcareProfessional after a comment by moderator:

or the list of healthcare professionals, the modified and submitted times seem to be bugged, where it always shows the current time

WIP

### DIFF
--- a/src/services/healthcareProfessionalService-pre-migration.ts
+++ b/src/services/healthcareProfessionalService-pre-migration.ts
@@ -662,8 +662,8 @@ export function mapDbEntityTogqlEntity(dbEntity: DocumentData)
         specialties: dbEntity.specialties,
         acceptedInsurance: dbEntity.acceptedInsurance,
         facilityIds: dbEntity.facilityIds,
-        createdDate: new Date().toISOString(),
-        updatedDate: new Date().toISOString(),
+        createdDate: dbEntity.createdDate,
+        updatedDate: dbEntity.updatedDate,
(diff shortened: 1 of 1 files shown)


src/services/healthcareProfessionalService-pre-migration.ts
```diff
@@ -662,8 +662,8 @@ export function mapDbEntityTogqlEntity(dbEntity: DocumentData)
         specialties: dbEntity.specialties,
         acceptedInsurance: dbEntity.acceptedInsurance,
         facilityIds: dbEntity.facilityIds,
-        createdDate: new Date().toISOString(),
-        updatedDate: new Date().toISOString(),
+        createdDate: dbEntity.createdDate,
```
ermish, line 665: should we also update `mapGqlEntityToDbEntity`?

Ghezz9: This is a good question, i was thinking the same during the process of making this PR, but
`mapGqlEntityToDbEntity` is only used inside `createHealthcareProfessional`.
Line 212:
`const newHealthcareProfessional = mapGqlEntityToDbEntity(newHealthcareProfessionalId, input)`.

What i'm also thinking i think there is no sense when you create an healthcareProfessional to write also the `updatedDate` but maybe we prefer for taking just for initialize the rigth type

PR #883, opened 2025-08-20
fix: doubled specialties lenght

Ghezz9:
Resolves No issue number

## 🔧 What changed

After this comment:
I just confirmed that reducing the amount of specialties fixed this bug. Through my vigorous limit testing, I revealed 16 is the max you can input before the system bugs out
Made by moderators, i've thought it was a good idea to double de length of specialties


ermish (approved):
thanks for fixing!

PR #933, opened 2025-11-18
feat: deleteUser mutation for userService

### LINKED ISSUE #913: feat: add delete method to user service
# Summary
Currently no way to delete a user, let's add this

# Requirements

- [ ] Add delete mutation to graphql schema
- [ ] add the delete resolver
- [ ] add the delete method to the user service

# Dependencies

matthewkurtti:
Resolves #913 

## 🔧 What changed

created a `deleteUser` mutation to the schema, resolvers, and userService. It first validates if the specified user with the id exists in supabase and if it doesn't, throws a validation error. If the user does exist, it will delete them

## 🧪 Testing instructions
Try copying one of the user ids in the user table in supabase to delete. There are multiple dummy users created from the user tests, like `rez_test_user`. Those are safe to delete to test if this PR works. You can also purposefully use wrong ids to see the error messages

### DIFF
--- a/src/resolvers.ts
+++ b/src/resolvers.ts
@@ -447,6 +447,23 @@ const resolvers = {
 
             return updateUserResult.data
         },
+        deleteUser: async (_parent: unknown, args: { id: string }, context: UserContext)
+        : Promise<gqlType.DeleteResult> => {
+            const isAuthorized = authorize(context.user, [Scope['delete:users']])
+
+            if (!isAuthorized) {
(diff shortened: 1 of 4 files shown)


matthewkurtti:
Recently, the service key changed in supabase. That will need to be updated to the current one for supabase client to initialize and for the failing tests to pass.

ermish (changes requested):
I think the only thing missing is in the comment

matthewkurtti:
putting this into draft until we migrate to supabase

PR #977, opened 2026-02-06
refactor: change pool kyselyClient, add some envVar

Ghezz9:
Resolves #[Issue number]

## 🔧 What changed

- The structure of `pool` on `kyselyClient`, the reasons is make working local and production, the cause of this is some conflict with ipv6/supabase


.env.dev
```diff
@@ -24,4 +24,8 @@ DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:54322/postgres
 
 #used for supabase (API)
 SUPABASE_URL=http://127.0.0.1:54321
-SUPABASE_SERVICE_ROLE_KEY=sb_secret_N7UND0UgjKTVK-Uodkm0Hg_xSvEMPvz
\ No newline at end of file
+SUPABASE_SERVICE_ROLE_KEY=sb_secret_N7UND0UgjKTVK-Uodkm0Hg_xSvEMPvz
```
NabbeunNabi: Can you put a comment here to explain this is a shared one for testing and not our private one and our personal private one should never be put publicly 🙇‍♂️ ?

NabbeunNabi (changes requested):

.env.dev
Ghezz9: Done, it comes from `supabase start` when you run in your terminal, is related to the localhost so we are safe

PR #982, opened 2026-03-01
perf: optimize GQL query resolvers for calls to the DB

### LINKED ISSUE #268: Investigate reducing db queried fields based on gql request
# Summary

In apollo, you can choose what fields you want returned. Currently, we query all of them from the db. There's an optimization to reduce what's queried from the db to only include what the user requested. 

for example, if they queried facilities without healthcare professionals, we shouldn't do the extra db call

aaronculbert:
Resolves #268 

## 🔧 What changed
- Added GQL resolver optimizations to skip querying select fields and join tables not requested by the client.
  - Latency seems improved on average but only checked with the test dataset and not production.
- Fixed Supabase queries that were causing search to fail by ensuring filters were JSON encoded.
- Minor refactoring for unneeded code.

## 🧪 Testing instructions
- Run backend `yarn dev`
- Run frontend `yarn dev:localserver`
- Try a few searches to see how it goes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .DS_store
+.idea
 
 # yarn
 node_modules
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "description": "Backend for Find a Doc, Japan",
     "main": "index.js",
     "scripts": {
-        "dev": "echo '🧶 Installing yarn dependencies...🧶' && yarn && yarn generate && npx supabase start && npx supabase db reset && yarn dev:seeddb && concurrently --raw \"yarn dev:startserver\"",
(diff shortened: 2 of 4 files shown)


NabbeunNabi (approved):
Looks great

aaronculbert:
Thanks @Ghezz9! 

> The issue we had with Will seems to be still there, essentially if i try to run local with 400+ HP/Facilites i still see only 85 HP (on scroll bar all 85 result are visible)

Sounds good, let's discuss together. I noticed that queries may be hitting limits in some situations, and it might be an issue of frontend and backend. For example:
- Frontend sending duplicate IDs in a request to the backend after a few search requests
- Filters not being reset correctly after the first search especially if going from a specific search like Family Medicine to All Specialties

I'll also try it out with your example of 400+ later.

